PR #14, opened 2016-11-24
Meeting minutes on 2016-11-24

unakatsuo:


Metallion:
I think it'd be a good idea to add a line on top that clearly says this is not current functionality but rather an idea of how it's expected to work in the future.

Metallion:
Well, that not current functionality line never came but the functionality came. :p Guess we can merge.

PR #16, opened 2016-11-24
Add RPMSpec VSCode extension.

unakatsuo:


Metallion:
Since you're the only VS user, I'll trust that this is how it should be. :) :+1:

PR #17, opened 2016-11-27
Fix unit test.

unakatsuo:
```
cd hypervisor
go test -v
```

got

```
# github.com/axsh/openvdc/hypervisor
./types_test.go:12: cannot use testDriver literal (type *testDriver) as type HypervisorDriver in return argument:
	*testDriver does not implement HypervisorDriver (missing CreateInstance method)
./types_test.go:27: cannot use testProvider literal (type *testProvider) as type HypervisorProvider in argument to RegisterProvider:
	*testProvider does not implement HypervisorProvider (missing SetName method)
FAIL	github.com/axsh/openvdc/hypervisor [build failed]
```

### DIFF
--- a/hypervisor/types.go
+++ b/hypervisor/types.go
@@ -9,7 +9,6 @@ import (
 type HypervisorProvider interface {
 	Name() string
 	CreateDriver() (HypervisorDriver, error)
-	SetName(string)
 }
 
 type HypervisorDriver interface {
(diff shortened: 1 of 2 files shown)


Metallion:
I haven't been able to test the LXC driver yet since it complains about an `lxc.pc` file not being found but that's probably a problem in my environment that I can fix later. :+1:

PR #21, opened 2016-11-28
Fix lxc test

b0r6:
Fixed the test.

### DIFF
--- a/hypervisor/lxc/lxc.go
+++ b/hypervisor/lxc/lxc.go
@@ -29,7 +29,7 @@ func (p *LXCHypervisorProvider) CreateDriver() (hypervisor.HypervisorDriver, err
 	return &LXCHypervisorDriver{
 		log:     log.WithField("hypervisor", "lxc"),
 		lxcpath: lxc.DefaultConfigPath(),
-		name:    currentLxcName,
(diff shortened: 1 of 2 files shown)


Metallion:
Getting this error:

```
[vagrant@openvdc-1box lxc]$ go test
INFO[0000] Registered hypervisor provider: lxc

INFO[0000] Creating lxc-container...                     hypervisor=lxc
lxc_container: utils.c: get_template_path: 1128 No such file or directory - bad template: download
lxc_container: lxccontainer.c: lxcapi_create: 1223 bad template: download
```

It seems to be looking for a nonexistant template called `download`.

b0r6:
The "bad template: download"-error seems to happen if "lxc-templates" isn't installed.

Metallion:
Note to self:

Steps required to make this test work:

```
sudo yum install lxc lxc-templates bridge-utils
sudo brctl addbr virbr0
```

Looks like the network settings are currently dependent on DHCP. Gonna fix it to use static networking instead.

Metallion:
Actually after more considering and discussion, we've decided to skip this test. A test that involves an actual container being started should be done at the integration test level and not unit test level.

PR #22, opened 2016-11-28
Generalize template

unakatsuo:
The original design of registry was to just store the set of machine image file references and VM specs. That results to simplify the command line for basic usage:

```
openvdc run centos/7/lxc
```

This PR intends to expand the idea to be able to support different type of resources not only for the VMs.

- [x] Update JSON schema
- [x] Apply new syntax to centos-7.json and move to ``templates/centos/7/lxc.json``
- [x] Add template for Null hypervisor.
- [x] ``openvdc register`` accepts the file path or remote URI to the plain .json file.
- [x] Bundle JSON schema and validation for templates.
- [x] ``openvdc template`` sub command.
- [x] registry/github.go downloads archive from master branch.


Metallion:
And since all branches are merging in the same commits, this one also needs the same merge conflict resolved.

Metallion:
:+1:

PR #24, opened 2016-11-28
Attributed offer from mesos-slave

unakatsuo:
Scheduler needs to know which hypervisor module is available on each mesos-slave node. So mesos-slave puts the supported hypervisor name to custom attribute then the scheduler reads it from offer.

To advertise that LXC is avaiable, run mesos-slave with``--attributes`` option:  

```
% mesos-slave --attributes="hypervisor:lxc"
```


Metallion:
This one needs to get its merge conflict resolved as well.

b0r6:
:+1:

PR #27, opened 2016-12-01
Place to define flags/switch variables for unit tests.

unakatsuo:
As example, defined "unittest.TestZkServer" variable to be available from *_test.go code in any namespaces.


Metallion:
Ran the unit tests manually on my environment for now and it works. :+1:

PR #28, opened 2016-12-01
Removed the echo line from the unit tests

Metallion:
This echo line at the end could ruin our day. Currently we're ok because there a `set -e` at the top of the script. That means the script exits immediately if the `go test` command exits with something other than 0.

However, if we ever remove that `set -e` in the future for whatever reason, the `echo` command's exit status will be used instead of `go test`'s and the CI will give false positives.

`go test`'s output is more than enough so I'll sleep better at night if we remove this line. :p


b0r6:
:+1:

Metallion:
Hmmm. Wonder what's up with this error in Jenkins.

```
java.lang.IllegalArgumentException: Null value not allowed as an environment variable: REPO_BASE_DIR
	at hudson.EnvVars.put(EnvVars.java:356)
	at hudson.model.StringParameterValue.buildEnvironment(StringParameterValue.java:56)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun.getEnvironment(WorkflowRun.java:366)
	at hudson.plugins.git.GitSCM.checkout(GitSCM.java:1075)
	at org.jenkinsci.plugins.workflow.steps.scm.SCMStep.checkout(SCMStep.java:109)
	at org.jenkinsci.plugins.workflow.cps.CpsScmFlowDefinition.create(CpsScmFlowDefinition.java:108)
	at org.jenkinsci.plugins.workflow.multibranch.SCMBinder.create(SCMBinder.java:85)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun.run(WorkflowRun.java:214)
	at hudson.model.ResourceController.execute(ResourceController.java:98)
	at hudson.model.Executor.run(Executor.java:404)
Finished: FAILURE
```

Metallion:
Looks like the CI is ok now. Merging.

PR #31, opened 2016-12-05
Instance state history

unakatsuo:
The instance state is a time series record so that the model class has to be extended to store that type.

- [x] CreateWithID() stores the last generated key to the sequence node.
- [x] FindLastKey() retrieves the last key from the sequence node.
- [x] Define InstanceState proto message.
- [x] Save created_at timestamp in [well-known protobuf type](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Timestamp).


Metallion:
:+1:

PR #33, opened 2016-12-07
Fix build script

unakatsuo:
- [x] schema.bindata.go should have the modtime from the last commit time.
- [x] Calucurate ``registry.GithubDefaultRef`` branch name from change history.
- [x] Checks on windows.


build.ps1
```diff
@@ -1,12 +1,32 @@
...
+} else {
+  # Found resource template/schema changes on this HEAD. Switch the default reference branch.
+  # Check if $GIT_BRANCH has something once in case of running in Jenkins.
+  $branch = if( $env:GIT_BRANCH ){
+      $env:GIT_BRANCH
+  }else{
+      $(git rev-parse --abbrev-ref HEAD)
+  }
+  $LDFLAGS="${LDFLAGS} -X 'registry.GithubDefaultRef=${branch}'"
+}
```
Metallion, line 29: Why go to github at all? Why not use what we have in the local source code? That just seems so much simpler to me

build.sh
```diff
@@ -15,13 +15,20 @@ $GOPATH/bin/govendor sync
...
+if (git rev-list origin/master | grep "${SCHEMA_LAST_COMMIT}") > /dev/null; then
+  # Found no changes for resource template/schema on HEAD.
+  # so that set preference to the master branch.
   LDFLAGS="${LDFLAGS} -X 'registry.GithubDefaultRef=master'"
 else
-  LDFLAGS="${LDFLAGS} -X 'registry.GithubDefaultRef=generalize-template'"
+  # Found resource template/schema changes on this HEAD. Switch the default reference branch.
+  # Check if $GIT_BRANCH has something once in case of running in Jenkins.
+  LDFLAGS="${LDFLAGS} -X 'registry.GithubDefaultRef=${GIT_BRANCH:-$(git rev-parse --abbrev-ref HEAD)}'"
 fi
```
Metallion, line 32: Same thing here. Isn't it way more simple to use the local source code when needed?

Metallion (changes requested):

build.ps1
unakatsuo: The functionality is to work the CLI syntax following:

```
openvdc run centos/7/lxc
```

The the template file is assumed to be on remote server in this case and it is expected to be on github as default. During development, the "templates/" folder will be changed so we want to set the default branch to a topic branch. For release build, it should be ok with the master branch.

Metallion: I think this kind of thinking will do more bad than good. It's much more user friendly to keep things simple and always default to the same templates which are kept on master.

To test other templates that are currently in development and on a remote server (ie a github branch) it should be enough to provide an absolute path/url.

Complicating this code will make it both error-prone and make it hard for users to know what the default templates repo is.

Metallion: Discussed IRL and I realize I was wrong. Merging now. :+1:

Metallion (approved):

PR #39, opened 2016-12-15
Trigger build on Appveyor

unakatsuo:


Metallion:
Why run on appveyor? Isn't it enough to run on our own CI?

Also I'm surprised you're adding something new on top of the CI 1box image. I thought we had agreed we were going to remove it from integration tests and re-purpose as development environment?

Metallion:
Been told irl that this is just to test if go build succeeds on Windows and upon closer inspection, I see that that's the only thing the appveyor script does.

I don't know why there's 1box integration related code in here but I'll merge. In any case I strongly suggest not spending more time on that 1box code because it will be removed very soon.

PR #40, opened 2016-12-16
Ci environment

toros11:
Scripts for building/destroying integration test environment with openvdc running across multiple nodes.

Functionality:
* Build and prepare mesos cluster.
* Store images as cache for each branch for reuse, base images can be rebuilt by giving a flag.
* Generate and copy-on-write images using the cache for boot up.
* Clean up after test has finished.


unakatsuo (approved):

Metallion (approved):
:+1:

PR #42, opened 2016-12-19
Multibox jenkins

toros11:
This integrates the multibox build environment into jenkins. 

* Added multibox node to the staging.
* Remove unnessecary sudo calls.
* Import variables from jenkins file in the config.


Metallion (approved):
yay :+1:

PR #53, opened 2016-12-22
Fix appveyor build.

unakatsuo:
It was due to ExecutionPolicy for sub .ps1 scripts.

```
.\build.ps1
The running command stopped because the preference variable "ErrorActionPreference" or common parameter is set to Stop: github.com/axsh/openvdc/vendor/github.com/Sirupsen/logrus
```


Metallion:
Why run zookeeper and then immediately shut it down again? Wasn't this job just for testing if compiling works on windows?

Metallion:
The CI fail is because the integration test server is currently down and needs maintenance. Unrelated to this PR request and can be skipped since the integration test is still underdevelopment anyway.

:+1: Thanks for fixing.

PR #55, opened 2017-01-04
Temporarily disable integration test

Metallion:
Currently there's a few issues with the integration test. Mainly the "waiting for ssh" part taking a very long time because DNS is enabled on the server side and the keypair setup isn't working quite right yet.

Since there are no actual integration tests written yet, it makes sense to skip the environment building part for now so it doesn't hold back other code. I will get to work on fixing this in another branch **immediately** since skipping tests is a big red flag.


akirapop:
+1

PR #37, opened 2016-12-14
Resource handler

unakatsuo:
- [x] Make "openvdc template *" work to check resource template file locally.
    - [x] Merge resource template default values and custom values from CLI options. 

- [x] "openvdc register" overwrite template parameters with JSON:

```bash
% openvdc register centos/7/lxc '{"interfaces": "macardd":"xxxx"}'
% openvdc register centos/7/lxc @file.json
% cat file.json | openvdc register centos/7/lxc
``` 
- [x] "openvdc run" works as same as "openvdc register".
- [x] "openvdc list" command from list-command branch.

Skipped Tasks:

- Overwrite parameters using command line option style.

-----

Example:

```bash
$ ./openvdc register centos/7/lxc '{"interfaces":[{"macaddr":"ff:ff:ff:ff:ff:ff", "ipv4addr":"192.168.10.10"}]}'
INFO[0000] Found template: centos/7/lxc                 
ID:"r-0000000040" resource:<id:"r-0000000040" template:<template_uri:"https://raw.githubusercontent.com/axsh/openvdc/resource-handler/centos/7/lxc.json" lxc:<lxc_image:<download_url:"https://images.linuxcontainers.org/1.0/images/d767cfe9a0df0b2213e28b39b61e8f79cb9b1e745eeed98c22bc5236f277309a/export" > interfaces:<macaddr:"ff:ff:ff:ff:ff:ff" ipv4addr:"192.168.10.10" > > > > 
$ ./openvdc show r-0000000040
ID: "r-0000000040"
resource: <
  id: "r-0000000040"
  template: <
    template_uri: "https://raw.githubusercontent.com/axsh/openvdc/resource-handler/centos/7/lxc.json"
    lxc: <
      lxc_image: <
        download_url: "https://images.linuxcontainers.org/1.0/images/d767cfe9a0df0b2213e28b39b61e8f79cb9b1e745eeed98c22bc5236f277309a/export"
      >
      interfaces: <
        macaddr: "ff:ff:ff:ff:ff:ff"
        ipv4addr: "192.168.10.10"
      >
    >
  >
>
```

PR #56, opened 2017-01-04
Resource handler without appveyor-fix

Metallion:
https://github.com/axsh/openvdc/pull/37 was ready for merge but then merged in https://github.com/axsh/openvdc/pull/53 which isn't ready for merge yet. This is the same PR as 37 but without 53 so we can merge it.


Metallion:
The appveyor failures are currently acceptable since they're unrelated to this PR and being worked on in https://github.com/axsh/openvdc/pull/53.

PR #47, opened 2016-12-19
Fix lxc

b0r6:
Lxc-containers can now be properly created, run, started, stopped and destroyed.

PR #57, opened 2017-01-04
Fix lxc without fix-appveyor

Metallion:
The same as https://github.com/axsh/openvdc/pull/47 but doesn't merge in https://github.com/axsh/openvdc/pull/53 which isn't ready for merge yet. Will fail the appveyor check but that is unrelated to this PR and being worked on in PR #53.


Metallion:
This can be merged since the appveyor problem has now been fixed on master and the unit test passed. The CI failure here is the integration test which is currently still under development and its server is currently down. :+1:

PR #58, opened 2017-01-05
Log command

b0r6:
Viewing the logs could be pretty annoying since the path keeps changing depending on the frameworkID and agentID. This command fetches logs via the master and then prints them.

For example; instead of manually navigating to this:
`/var/lib/mesos/agent/slaves/76f4cb0f-51e6-4afd-832c-7a123c109c47-S3/frameworks/967e9537-91d2-4e7f-bfce-175fab1506e4-0026/executors/vdc-hypervisor-null/runs/752452f6-66db-4906-9a41-341382fce361/stderr`

You'd use this command instead:
`./openvdc log i-0000000001`


cmd/openvdc/cmd/root.go
```diff
@@ -49,6 +50,7 @@ func init() {
 	// Cobra also supports local flags, which will only run
 	// when this action is called directly.
 	RootCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
-	RootCmd.PersistentFlags().StringVarP(&util.ServerAddr, "server", "s", "localhost:5000", "gRPC API server address")
+	RootCmd.PersistentFlags().StringVarP(&util.MesosMasterAddr, "master", "", "127.0.0.1:5050", "Mesos Master node address")
+	RootCmd.PersistentFlags().StringVarP(&util.ServerAddr, "server", "s", "127.0.0.1:5000", "gRPC API server address")
```
Metallion, line 54: Are these IP addresses configurable? The integration tests will soon require them to be. :) Can be handled in a different PR if not though.

cmd/openvdc/cmd/log.go
```diff
@@ -0,0 +1,71 @@
...
+		}
+
+		instanceID := "VDC_" + args[0]
+
+		split := strings.Split(util.MesosMasterAddr, ":")
+		mesosMasterAddr := split[0]
+		mesosMasterPort, err := strconv.Atoi(split[1])
+
+		if err != nil {
+			log.Errorln("Couldn't convert string to int") //  <--- lol
```
Metallion: This should include the actual message from `err`. Same goes for the logs further down this code.

I also quickly tried to look up the `log.Errorln` function but couldn't find it right away. Does it exit with a non-zero exit code? The way it's named, I'd expect it to log the error and then continue executing the code below but this type of error strikes me as something that should be fatal. Please don't hesitate to tell me if I'm wrong. :)

cmd/openvdc/cmd/root.go
b0r6: Yeah these will soon be handled in some better way. As for now, these can be passed as flags when sending the command though.

Metallion: I see. Thanks. :)

cmd/openvdc/cmd/log.go
b0r6: Ah yeah, you're right. I'll fix this.

Metallion:
Looks good to me. :+1:

PR #59, opened 2017-01-05
Some multibox fixes

Metallion:
* Server side DNS lookups were enabled on all CI nodes which causes very long delays when DNS isn't available. Disabled those which a quick hack and should be done in the seed image in the future.
* All of multibox's data was in the `/data` directory which is part of the root partition and has only 50 GB available. This will fill up very quickly so I moved it to `/data2` which is on a separate 1TB harddrive that was not used at all at the moment.
* A couple of bugfixes related to SSH keypair generation

* Re-enabled the integration test since it should now be working... It's environment building part at least. The actual tests still need to be written and that's next on my plate.


b0r6:
Looks good. :+1:

PR #63, opened 2017-01-06
Just a quick fix

b0r6:
A few branches, including the master, are still referring to the branch "resource-handler" which doesn't exist anymore. So here's a quick solution to that.

### DIFF
--- a/registry/github.go
+++ b/registry/github.go
@@ -24,7 +24,7 @@ const (
 )
 
 // Build time flag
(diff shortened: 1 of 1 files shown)


Metallion:
:+1: 

The CI failure is related to integration test environment building and the unit tests passed. I will investigate the integration test environment failure but it shouldn't stop this PR from being merged.

PR #64, opened 2017-01-07
Load user config from $HOME/.openvdc

unakatsuo:
``openvdc`` CLI has several common flags, that spf13/cobra says PersistentFlags. They had to be passed to the CLI everytime in order to specify the server address to connect. They should be able to save as the user local configuration file.

``github.com/spf13/viper`` provides the feature to load configuration from multiple sources and to get merged value in loading priority.

Discussions:

1. ``spf13/viper`` supports multiple formats, json, yaml, toml and hcl. which is/are better to support? Or which format should be rejected.
    - My opinion is ``.toml`` and then ``.json`` or ``.yaml``. ``.hcl`` can be ignored.
2. Might be out of scope this PR, ``openvdc-scheduler`` should it use the same framework?

TODO:

- [x] non-default config file can be set with ``--config`` flag.
- [x] Binds ``--server`` and ``--master`` flags to viper namespace. 
- [x] Install default configuration file ``config.toml`` if not exist.
- [x] On windows, ``openvdc help`` shows correct user home location.

### DIFF
--- a/cmd/openvdc/cmd/log.go
+++ b/cmd/openvdc/cmd/log.go
@@ -7,8 +7,8 @@ import (
 
 	mlog "github.com/ContainX/go-mesoslog/mesoslog"
 	log "github.com/Sirupsen/logrus"
-	"github.com/axsh/openvdc/cmd/openvdc/internal/util"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 var tail bool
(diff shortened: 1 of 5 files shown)


unakatsuo:
```
PS > .\openvdc help
Usage:
  openvdc [command]

Available Commands:
  console     Connect to an instance
  create      Create an instance from resource
  destroy     Destroy an instance
  list        List instances
  log         Print logs of an instance
  register    Register new resource.
  run         Run an instance
  show        Show a resource
  start       Start an instance
  stop        Stop a running instance
  template    Operations for resource template
  unregister  Unregister a resource
  version     Show version

Flags:
      --config string   Load config file from the path (default "C:\Users\katsuo\.openvdc\config.toml")
      --master string   Mesos Master node address (default "127.0.0.1:5050")
      --server string   gRPC API server address (default "127.0.0.1:5000")
  -t, --toggle          Help message for toggle

Use "openvdc [command] --help" for more information about a command.
```

Metallion:
:+1: CI failure is due to a known issue in the integration test building code unrelated to this PR and is currently being worked on in another branch. Unit tests were green.

PR #67, opened 2017-01-10
CI merge master locally fix

Metallion:
The way merging before build was implemented introduced a problem where the RELEASE_SUFFIX would hold a different commit than the one on github. This also broke the OpenVDC templates repository.

This PR fixes the former and maybe fixes the latter. Whether or not the latter is really fixed will be shown by the integration test soon.


b0r6:
:+1:

PR #68, opened 2017-01-11
Set timestamp to all models and model class improvement.

unakatsuo:
Added ``created_at`` timestamp field to all structs in ``model.proto``. Then we became to need the feature in model class to facilitate filling values for these well known field.

- [x] Add ``created_at`` field.
- ``backend.ProtoModelBackend`` wrapper: 
    - [x] Accepts proto.Message object and marshal inside: ``Create()``, ``CreateWithID()``, ``Update()``
    - [x] Filter mechanism that finds the field and applies the value.
- [x] Create TimestampFilter.
- [x] Enable the backend wrapper to all ``model/*.go``.


b0r6:
This looks good as well. Nothing to add. +1

PR #71, opened 2017-01-11
Multibox changes and OpenVDC install

Metallion:
This PR does a couple of things

* Renamed `kill.sh` and `destroy.sh` to `destroy_leaving_cache.sh` and `destroy_with_cache.sh` to better reflect what they do
* Install openvdc packages in the correct nodes so we can finally start writing tests
* Removed wget dependency and use curl instead
* Some refactoring that allows us to use SSH outside of the ind-steps loop

* Temporarily disabled the feature that copies from develop if rebuild flag isn't set. Will re-enable this once the CI is changed to be kicked off manually with the rebuild flag. Gonna write a simple test first though.

Next step that I'll start working on immediately: The actual tests that will run after OpenVDC was installed.


unakatsuo:
can we merge [teardown-ci-multi](/axsh/openvdc/compare/teardown-ci-multi) branch into this PR?

Metallion:
I would rather get this merged into master asap. I will ask @toros11 to review it once he gets in today. Should be able to merge it today then.

toros11:
+1

21859a7d1d43, this code was not intended to detect whether to rebuild or not. It was meant to be used in code that would detect where to clone from, but this is incomplete for now so go ahead and merge.

Metallion:
I see. Thanks for the review. :)

PR #72, opened 2017-01-11
Clean a binary file added wrongly.

unakatsuo:
I added boltdb binary file which is generated from unit test.

Caused by #68.


Metallion:
The CI failure is again that known issue where a previous build failure doesn't clean up after itself. I will fix it very soon.

This PR can go through as everything but that passed. :+1:

PR #73, opened 2017-01-11
Protoc go generate

unakatsuo:
``//go:generate ...`` magic comment allows to run code generator.  Let's run them via "go generate" and make the build steps simpler. 

[The Go Blog: Generating code](https://blog.golang.org/generate)


Metallion:
:+1:

PR #75, opened 2017-01-12
build: New EPEL release: 7.8 -> 7.9

unakatsuo:


Metallion:
I think in multibox we should revert back to `yum install -y epel-release` since we don't require the latest docker there and this problem is going to repeat itself every time a new version is released.

For places where the latest docker is required I would actually suggest using Fedora instead of Centos or find other means of installing docker.

In any case that can be done in another PR so this can be merged. :+1:

PR #77, opened 2017-01-13
Ensure to run the teardown steps for ci-multi cluster

unakatsuo:
Our integration test was failing sometime due to the garbage from the previous build.  ``Cleanup Environment`` Jenkins stage should run all the steps even if they met errors otherwise virtual machines for testing or network bridge remain on the server.

By wrapping the destroy script in ``finally`` block, the ci-multi cluster is wiped out when the building/testing job was canceled.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -75,11 +75,14 @@ def stage_integration(label) {
     stage "Build Integration Environment"
     write_build_env(label)
 
-    sh "cd ci/multibox/ ; ./build.sh"
-    stage "Run Tntegration Test"
-    // This is where the integration test will be run
-    stage "Cleanup Environment"
-    sh "cd ci/multibox/ ; ./destroy.sh --kill"
+    try {
+      sh "cd ci/multibox/ ; ./build.sh"
+      stage "Run Tntegration Test"
+      // This is where the integration test will be run
+    } finally {
+      stage "Cleanup Environment"
+      sh "cd ci/multibox/ ; ./destroy.sh --kill"
+    }
   }
 }
 
--- a/ci/multibox/config.source
+++ b/ci/multibox/config.source
@@ -1,3 +1,4 @@
+#!/bin/bash
 # The nodes to build. If any of these are replaced by a physical device,
 # comment it out here.
 NODES=(
--- a/ci/multibox/ind-steps/common.source
+++ b/ci/multibox/ind-steps/common.source
@@ -60,7 +60,7 @@ function stop_masquerade() {
     [ "$?" != "0" ]
     $skip_step_if_already_done; set +e
     sudo iptables -t nat -D POSTROUTING -s "${subnet}" -j MASQUERADE
-  )
+  ) # ; prev_cmd_failed ### TODO: bashstep missing feature (axsh/openvdc#77)
 }
 
 function run_steps () {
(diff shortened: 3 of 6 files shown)


Metallion:
Why are you removing the `$prev_cmd_failed`'s?

unakatsuo:
That stops the job at the middle. The case is that VMs are removed but bridge remains because the deleting step appears after stopping VMs.

triggers:
If the "do" part of a step fails, the whole script should exit.  Everything is simple, so you can make exceptions if it really helps get the job done.  In that case, a comment should be written explaining why breaking the usual pattern is the best way to do things. 

In general, the fact that one step precedes another implies a dependency, so in general later steps should not be attempted.  If there is no dependency, maybe the whole script should be broken into two smaller build scripts?  Anyway, probably the steps should be refactored so they make sense, and all the prev_cmd_failed things should stay.

triggers:
Looking at the code now...
For example, I think what @unakatsuo is saying it that it is OK for

            sudo iptables -t nat -D POSTROUTING -s "${subnet}" -j MASQUERADE

to fail.  In that case, do

            sudo iptables -t nat -D POSTROUTING -s "${subnet}" -j MASQUERADE || true

And leave the prev_cmd_failed in there.  That way, we can say the step "succeeds" even though the iptables command returns an error.  ((Just guessing based on a quick look at the code and the comments))

unakatsuo:
I don't want to put ``|| true`` for all the commands in teardown step. What ``prev_cmd_failed`` actually does is ``reportfailed`` and ``exit 255``. Error reporting might be worth in this use case so the change I can propose is that:

1.  ``prev_cmd_failed`` takes a switch to skip ``exit 255``.
2. Prepare another function, such like ``prev_cmd_failed_skip``, that calls ``reportfailed`` and continue.
3. Allow ``prev_cmd_failed`` not mandate as my change.
4. Allow ``) ; prev_cmd_failed || :``.

triggers:
Because you want the error message?   How about defining a function in the file like:

     reportfail_continue() { echo "Command failed ($*)" ; return 0 }

and then

      sudo iptables -t nat -D POSTROUTING -s "${subnet}" -j MASQUERADE || reportfail_continue

Maybe your proposal (or something similar)  makes sense.  Still thinking about it. You can prototype by writing functions local to the file and we can see how it looks.

unakatsuo:
Added 4th idea.

triggers:
|| : will not work, because it uses exit   :-(

Another alternative is with comments, but this will not report anything to stderr, if that is what you want.  Really this is the simplest short term solution.

      ) #  prev_cmd_failed ### no steps below depends on this step

unakatsuo:
```
    # for consistency, start to use the variable form for everything
    : ${prev_cmd_failed:='eval [ $? = 0 ] || exit 255'}
```
The variable version does not do ``reportfailed()``. I think ``prev_cmd_failed`` is not suitable to place here to expect to see report something.

I'll keep the end of ``)`` as blank where the steps needs to ignore errors. Please approve if you don't have any concerns out of this topic.

triggers:
>  The variable version does not do reportfailed(). 

That is a bug...still not 100% sure what is best here, but the variable version should match the function version :-(

> I'll keep the end of ) as blank where the steps needs to ignore errors. 

It needs to be commented.  Typically that would be a bug, and somebody editing the code later will probably put prev_cmd_failed back in if they don't know that it was removed for a specific reason.

When there is time, we need to have a meeting.  There may be interesting issues being raised for this use case.

Metallion:
Merging this for now. As said by @triggers we'll have a meeting later to decide how exactly we're going to restructure this code. :+1:

PR #80, opened 2017-01-19
Reboot command

unakatsuo:
This adds the possibility to reboot containers via this command ./openvdc reboot i-xxxxxx
Also added a new instance state which the containers are given during the reboot.


b0r6:
Yeah so this just adds the possibility to reboot containers via this command `./openvdc reboot i-xxxxxx`

Oh and rebooting an instance should probably cause its state to get updated as well I guess, which it doesn't at the moment.

If so, should it use the existing states and go from `running > stopping > stopped > starting > running`
or would it be better to add a new one called "rebooting"?

unakatsuo:
Generally, each operation should have unique state. ``rebooting`` is the right choice.

b0r6:
Added a new instance state for when they're rebooting.

unakatsuo (commented):
Please add descriptions to this PR.

Metallion:
@unakatsuo I think his description is in his comment above.

@b0r6 Since the acceptance test code is now merged, could you write a test case for this feature?

ci/acceptance-test/tests/cmd_reboot_test.go
```diff
@@ -0,0 +1,34 @@
...
+		if err !=nil {
+			return err
+		}
+		result := gjson.GetBytes(buf, "instance.lastState.state")
+		if result.String() == "RUNNING" {
+			return nil
+		}
+		time.Sleep(5 * time.Second)
+		return WaitContinue
+	})
```
Metallion: The same code is found here. https://github.com/axsh/openvdc/blob/master/ci/acceptance-test/tests/cmd_console_test.go#L18-L30

It should be refactored into a function and re-used.

ci/acceptance-test/tests/cmd_reboot_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	RunCmdAndReportFail(t, "openvdc", "reboot", instance_id)
```
Metallion: Just running the command and checking the exit status isn't enough. A test needs to be done to check if the instance really has rebooted.

A simple way to do this would be to add a line to `rc.local` that creates a file. Then after reboot, check if the file exists.

Metallion (changes requested):

Metallion (approved):

PR #81, opened 2017-01-19
Rpm cleanup

akirapop:
This code is ready to merged into master. The cron job on the ci is scheduled to run daily at 26:00 (02:00 am).


Jenkinsfile
```diff
@@ -104,7 +114,8 @@ if( buildParams.BUILD_OS != "all" ){
...
 // Using .each{} hits "a CPS-transformed closure is not yet supported (JENKINS-26481)"
 for( label in build_nodes) {
-  stage_unit_test(label)
-  stage_rpmbuild(label)
-  stage_integration(label)
+//stage_unit_test(label)
+//stage_rpmbuild(label)
+//stage_integration(label)
+  stage_clean_rpms(label)
 }
```
unakatsuo, line 108: Please revert the changes to ``Jenkinsfile``. This causes the existing build process halts and cleanup tasks is performed by another Jenkins job not by ``citest`` job.

Jenkinsfile.garbage_collection
```diff
@@ -0,0 +1,85 @@
...
+    }catch(org.jenkinsci.plugins.workflow.steps.FlowInterruptedException err) {
+      // Only ignore errors for timeout.
+    }
+    checkout scm
+//  checkout_and_merge()
+    // http://stackoverflow.com/questions/36507410/is-it-possible-to-capture-the-stdout-from-the-sh-dsl-command-in-the-pipeline
+    // https://issues.jenkins-ci.org/browse/JENKINS-26133
+    RELEASE_SUFFIX=sh(returnStdout: true, script: "./deployment/packagebuild/gen-dev-build-tag.sh").trim()
+}
+
```
unakatsuo: Please cleanup the code. I think the lines of code by the line here has no meaning for cleanup job.

repo_cleanups/remove_stale_rpm.sh
```diff
@@ -0,0 +1,231 @@
+#!/bin/bash
```
unakatsuo: This file should be moved under ``ci/``.

unakatsuo (changes requested):
I found the bunch of  lines which should be removed. Please fix.

Jenkinsfile.garbage_collection
akirapop: Jenkinsfile.garbage_collection now has only code directly related to the repo. cleanup

repo_cleanups/remove_stale_rpm.sh
akirapop: Ok, I've moved repo_cleanups/ to this directory.

Jenkinsfile
akirapop: Return the Jenkinsfile to its orginal, is that what you mean? Jenkinsfileは、元に戻すっていうこと、でしょう？

unakatsuo: Why did you change this file? ``Jenkinsfile.garbage_collection``, isn't it the script file to perform cleanup job?

akirapop: Do you mean, why did I move the file? I just thought it made more sense to reside in the ci/repo_cleanups directory, since that is what it is logically connected to. The jenkins configuration was updated to reflect this new location.

unakatsuo: I mean why this branch has the change to the root ``Jenkinsfile``.  I understand that why ``Jenkinsfile.garbage_collection`` is added and there is new Jenkins job ``garbage_collection`` refers to ``Jenkinsfile.garbage_collection``.

I estimate that you applied changes to the root ``Jenkinsfile`` at the beginning and later copied the modified ``Jenkinsfile`` to ``Jenkinsfile.garbage_collection`` at af29f96eb2b24. But your change still remain on the root ``Jenkinsfile``.

https://github.com/axsh/openvdc/tree/d6c8cea3e7fc742733c3cb4dba61bd6d272e2e64

akirapop: I commented out the stage_xxx steps because they are unrelated to what this branch does. I think it is a waste of resources to have Jenkins run builds and tests on code not relevant to this branch.

unakatsuo: So, this branch is going to be merged. What will happen?

I don't complain about the series of changes to the file during development meantime. Why are the changes still appearing at the HEAD of this branch?

Metallion: I agree with @unakatsuo on this one.

It's true that the main Jenkinsfile runs tests generally unrelated to this branch but if it gets merged with those lines commented out, it will stop running the tests on master (and future branches) too. It's ok to comment these out during development but before merge, it should be restored back to how it was.

Also in general I think it's good to run tests at least once before merging even if unrelated because you just never know. Some times unrelated code can have unexpected consequences.

Jenkinsfile.garbage_collection
```diff
@@ -0,0 +1,85 @@
...
+BUILD_OS=${label}
+REBUILD=${buildParams.REBUILD}
+RELEASE_SUFFIX=${RELEASE_SUFFIX}
+# https://issues.jenkins-ci.org/browse/JENKINS-30252
+GIT_BRANCH=${env.BRANCH_NAME}
+BRANCH_NAME=${env.BRANCH_NAME}
+BRANCH=${env.BRANCH_NAME}
+"""
+  writeFile(file: "build.env", text: build_env)
+}
```
Metallion: ~~I haven't read the actual cleanup script yet but this looks like code copy-pasted from the main Jenkinsfile. I think there are many variables here that are not used. For example there is `LEAVE_CONTAINER` but I don't think this job starts any containers.~~

~~To keep the code clean and readable, any unused variables should be removed.~~

I see below that a copy of this file was committed with the code cleaned up. I guess you just forgot to call `git rm` on this.

ci/repo_cleanups/Jenkinsfile.garbage_collection
```diff
@@ -0,0 +1,26 @@
+#!groovy
+
+import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException
+// http://stackoverflow.com/questions/37425064/how-to-use-environment-variables-in-a-groovy-function-using-a-jenkinsfile
+import groovy.transform.Field
+
+def checkout_and_merge() {
+    checkout scm
+    sh "git merge origin/master"
+}
```
Metallion: This function is not used (and isn't needed in the garbage collection job) so it should be removed.

ci/repo_cleanups/Jenkinsfile.garbage_collection
```diff
@@ -0,0 +1,26 @@
...
+    stage "RPM Repository Cleanups"
+    checkout scm
+    sh "git remote prune origin"
+    sh "cd ci/repo_cleanups ; ./remove_stale_rpm.sh"
+  }
+}
+
+node() {
+    stage "Checkout"
+//  checkout_and_merge()
```
Metallion: I think it's better to remove commented out code before merging. Git makes it possible to get old code back and I think it makes the code cleaner.

Jenkinsfile.garbage_collection
Metallion: As mentioned above, you seem to have forgotten to remove this file when you moved it to `ci/repo_cleanups`.

You can use `git rm` followed by `git commit` to remove it.

ci/repo_cleanups/Jenkinsfile.garbage_collection
```diff
@@ -0,0 +1,26 @@
...
+    stage_clean_rpms()
+}
+
```
Metallion: Other than those two comments, it's a clean file now. ^_^ Very good.

Metallion (changes requested):
One more general comment I have is the commit messages.  They should be a little more descriptive of the change you made.

Here's a couple PRs with examples of good commit messages.

https://github.com/axsh/openvdc/pull/68
https://github.com/axsh/openvdc/pull/58

No need to go back and change commit messages but just something to keep in mind for the future. ;)

repo_cleanups/remove_stale_rpm.sh
Metallion: It looks like you forgot to call `git rm` here too. For the future, `git mv` is a useful command to rename files with git being aware of it. `git mv foo bar` is basically the same as:

```
mv foo bar
git add bar
git rm foo
```

ci/repo_cleanups/Jenkinsfile.garbage_collection
akirapop: I have removed this comment.

ci/repo_cleanups/Jenkinsfile.garbage_collection
akirapop: It has been removed.

akirapop: Likewise, Jenkinsfile.garbage_collection has been removed from the top of the tree for this branch.

Jenkinsfile.garbage_collection
akirapop: As commented below, the file has been removed.

Jenkinsfile
akirapop: OK, I've replaced this Jenkinsfile with the current version in master.

ci/repo_cleanups/remove_stale_rpm.sh
```diff
@@ -0,0 +1,231 @@
...
+
+}
+
+#########################################################################################
+git_branches="$(run_git)"
+
+script=$(write_script)
+
+## Send the script generated above to the dh machine
+echo -e "${script/_STUB_/${git_branches}/}" | $SSH_REMOTE /usr/bin/bash
```
Metallion: Writing the whole script in a function is not a good idea I think. For starters it removes syntax highlighting and it's likely to have other side effects too.

If it is required to send the script to another machine to execute it there, it's better to save the script normally as its own file and then send that through SSH either directly from the Jenkinsfile or from a wrapper script.

ci/repo_cleanups/remove_stale_rpm.sh
```diff
@@ -0,0 +1,231 @@
+#!/bin/bash 
+
+
+function run_git {
+#   git for-each-ref --sort=-committerdate refs/remotes --format='%(HEAD)%(refname:short), %(committerdate:relative)'
```
Metallion: As mentioned above I think it's cleaner to remove old code. They're still preserved in git history if it's needed to bring them back later.

ci/repo_cleanups/remove_stale_rpm.sh
```diff
@@ -0,0 +1,231 @@
...
+        rpmdate=${dt:0:8}     # yyyymmdd is the format
+
+        ndays=$(delta_dates ${rpmdate} ${now})
+
+        if [[ "${dt}" = "${current}" ]]; then
+            continue
+        fi
+
+        if [[ ${ndays} -gt 14 ]]; then
+            ## Just in case the continue command does not work as I think it should...
```
Metallion: What is the concern you have with the continue command? The way it works is documented well in the [bash man page](https://linux.die.net/man/1/bash).

> continue [n]
Resume the next iteration of the enclosing for, while, until, or select loop. If n is specified, resume at the nth enclosing loop. n must be ≥ 1. If n is greater than the number of enclosing loops, the last enclosing loop (the ''top-level'' loop) is resumed. When continue is executed inside of loop, the return value is non-zero when n is ≤ 0; Otherwise, continue returns 0 value. When continue is executed outside of loop, the return value is 0.

I also used to write code like this but I think it's better to spend a little extra time and understand exactly what the commands you use are doing. That way there's no need for clauses like this and it simplifies the code.

ci/repo_cleanups/remove_stale_rpm.sh
```diff
@@ -0,0 +1,231 @@
...
+#   git for-each-ref --sort=-committerdate refs/remotes --format='%(HEAD)%(refname:short), %(committerdate:relative)'
+    git for-each-ref --sort=-committerdate refs/remotes --format='%(refname:short), %(committerdate:relative)'
+}
+
+
+function write_script {
+
+    var=$(cat <<"EOF"
+
+    declare -A  git_branches    ## Store the list of git branches in a hash
```
Metallion: I wonder if there is a way to do it without using hashes. They require bash version 4 and it's a complicated data structure which is not what bash is generally good at.

I haven't understood the script enough yet to provide a good alternative but I'll keep reading a bit more.

ci/repo_cleanups/remove_stale_rpm.sh
```diff
@@ -0,0 +1,231 @@
...
+            echo "!! Not in the proper directory! Should be in \"${rpm_repo_dir}\"..."
+            exit 1
+        fi
+        
+        for no_rm in "master" "develop"; do
+           if [[ "${dead_dir}" = "${no_rm}" ]]; then
+               echo "Cannot remove \"${no_rm}\". Ignoring."
+               return 0
+           fi
+        done
```
Metallion: Still haven't understood the script completely but relying on what the current working directory is seems a bit of a red flag to me. I would prefer constructing the absolute path to be deleted. We already have the `rpm_repo_dir` variable so it shouldn't be too hard to do.

I plan to read some more to see what exactly is being done and then I'll give more concrete suggestions. :)

unakatsuo (approved):

ci/repo_cleanups/remove_stale_rpm.sh
```diff
@@ -0,0 +1,231 @@
...
+    
+        echo "rm -rf ./${dead_dir} "
+        rm -rf ./${dead_dir}
+        if [ $? -ne 0 ]; then
+           echo "       rm FAILED for ./${dead_dir} "
+        fi
```
Metallion: Instead of error checking, it's better to put `set -xe` at the top of your script.

The `-e` flag will cause your script to fail and stop execution if any of its subprocesses (such as `rm`) exit with a status other than 0.

The `-x` flag will print the lines currently executed to stdout.

Metallion (changes requested):
Still reviewing the script but I already placed a couple comments that I wanted to publish. :)

ci/garbage_collection/repo_cleanups/remove_stale_rpm.sh
```diff
@@ -0,0 +1,229 @@
+#!/bin/bash
```
Metallion: This file still has the old code and isn't called anywhere. I guess you forgot to run `git rm`?

Jenkinsfile
```diff
@@ -38,16 +38,18 @@ RELEASE_SUFFIX=${RELEASE_SUFFIX}
 GIT_BRANCH=${env.BRANCH_NAME}
```
Metallion, line 38: The root Jenkinsfile should be made the same as the current master branch. It seems you have copied over the Jenkinsfile from your local master branch but that one was not up-to-date with the master branch on github. Here's how I suggest we fix it.

In the openvdc root directory:

```
# Bring your local master up-to-date
git checkout master
git pull --rebase

# Merge updated local master into your branch. This will cause a merge conflict on Jenkinsfile.
git checkout rpm_cleanup
git merge master

# Get the Jenkinsfile from current master on github.
curl https://raw.githubusercontent.com/axsh/openvdc/master/Jenkinsfile -o Jenkinfile

# Solve merge conflict.
git add Jenkinsfile
git commit
```

Metallion (changes requested):
I see this output in the CI:

```
Checking for stale rpm repos under master...
rm -rf /var/www/html/openvdc-repos/20170110092208git66d1ba5  (command not executed)
rm -rf /var/www/html/openvdc-repos/20170111053806git384446e  (command not executed)
rm -rf /var/www/html/openvdc-repos/20170111072135gite7faf4c  (command not executed)
```

It checks for old commits under `/var/www/html/openvdc-repos/` but the master repos are under `/var/www/html/openvdc-repos/master/`.

akirapop:
I have made the following changes to the code:  I have factored out the portions of the code that determine which directories (which are mapped to git branch names) need to be removed: Does the git branch no longer exist? yes - Remove. If "no" - has the branch been pushed to within 14 days? No? Remove. This code can now be incorporated also into the cache_cleaning code. 

I've changed the directory structure:
     openvdc/ci/repo_cleanups/garbage_collection_misc.sh
                                              /prune_branches.sh
                                              /repo_cleanups/
                                                                       /rpm_garbage_collection.sh

<on the cache_cleanup branch>
                                              /cache_cleanups/
                                                                        /cache_garbage_collection.sh

akirapop:
Other changes:

   (1) The script runs directly on the dh now, so there is no writing the script to a sh variable and then
         running that via ssh

   (2) No hash is used to find directories that need to be pruned. Instead, the comm command
        (file comparison command) is used.

Metallion (approved):

PR #87, opened 2017-01-20
Improve scheduler command options: --zk, --master

unakatsuo:
--zk and --master options are extended to support ``zk://`` URL syntax similar to the mesos --zk option. The URL allows to specify multiple zookeeper hosts.

```
zk://host1:port1,host2:port2/base
```

For --zk option, let go-zookeeper recover from a zookeeper node failure. 
For --master option, mesos-go detects mesos-master node endpoint from the zookeeper catalogue. It is required for the mesos multi master configuration.

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -4,6 +4,7 @@ import (
 	"net"
 
 	"github.com/axsh/openvdc/model"
+	"github.com/axsh/openvdc/model/backend"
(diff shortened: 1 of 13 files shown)


Metallion:
I didn't test it myself but I see tests were written so I'm going to trust those. Looks good to me. :+1:

PR #91, opened 2017-01-27
Executor local configuration file.

unakatsuo:
Support executor's local configuration file. 

```
/etc/openvdc/executor.toml
```

``--config`` option loads the conf file from another location.

```
openvdc-executor --config=/var/tmp/custom.toml
```


Metallion:
Looks like merging https://github.com/axsh/openvdc/pull/87 introduced a conflict here.

Metallion (changes requested):
* A default config file should be installed by the openvdc-executor rpm.

* Actually @b0r6 was also doing something like this in his [lxc-network](https://github.com/axsh/openvdc/compare/lxc-networking#diff-952f06bd8fa744cb29dc433d450563f4R365) branch.

At first sight this implementation looks cleaner but might be a good idea to quickly get on the same page with him.

unakatsuo:
Will talk to @b0r6 once.

unakatsuo:
As discussed with @b0r6, the conf file change in ``lxc-network`` will be reverted. This branch will be left.

Metallion (approved):

Metallion:
Merging the scheduler config file caused some merge conflicts here. I solved them and will merge if the CI passes. Most likely it will.

PR #94, opened 2017-01-27
Scheduler local configuration file

unakatsuo:
Support scheduler's local configuration file. 

```
/etc/openvdc/scheduler.toml
```

``--config`` option loads the conf file from another location.

```
openvdc-scheduler--config=/var/tmp/custom.toml
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -41,6 +41,6 @@ else
 fi
 
 go build -ldflags "$LDFLAGS" -v ./cmd/openvdc
-go build -ldflags "$LDFLAGS" -v ./cmd/openvdc-scheduler
+go build -ldflags "$LDFLAGS -X 'main.DefaultConfPath=/etc/openvdc/scheduler.conf'" -v ./cmd/openvdc-scheduler
 go build -ldflags "$LDFLAGS" -v ./cmd/openvdc-executor
 echo "Done"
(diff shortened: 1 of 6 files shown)


Metallion (changes requested):
Same comment as https://github.com/axsh/openvdc/pull/91. The `openvdc-scheduler` rpm package needs to install a default config file.

build.sh
```diff
@@ -41,6 +41,6 @@ else
 fi
 
 go build -ldflags "$LDFLAGS" -v ./cmd/openvdc
-go build -ldflags "$LDFLAGS" -v ./cmd/openvdc-scheduler
+go build -ldflags "$LDFLAGS -X 'main.DefaultConfPath=/etc/openvdc/scheduler.conf'" -v ./cmd/openvdc-scheduler
```
Metallion: Why the .conf extension? In other places we're going with `toml` files so I think we should do it here too. I know viper supports multiple extensions and that's fine but let's use .toml for the default file.

Metallion:
We should remove the `pkg/rhel/sysconfig-openvdc` file since it is no longer used. Other than that I think it's good.

unakatsuo:
I have merged ``scheduler-config-file``. ``pkg/rhel/sysconfig-openvdc`` no longer exists. Applied same change for #91 as well.

Metallion (approved):

Issue #89, opened 2017-01-25
Acceptance test re-use master cache


Metallion:
### Problem

There was code in place to use the acceptance test cache for the master branch when building a new branch and the REBUILD variable isn't set.

I temporarily disabled that code because I wrongly assumed that we couldn't set the REBUILD variable manually while the CI is still kicked off automatically

### Solution

Re-enable that code.

Metallion:
Done.

PR #96, opened 2017-01-30
Reinstate master cache reuse

Metallion:
Solves https://github.com/axsh/openvdc/issues/89. Copy paste from that issue:

### Problem

There was code in place to use the acceptance test cache for the master branch when building a new branch and the REBUILD variable isn't set.

I temporarily disabled that code because I wrongly assumed that we couldn't set the REBUILD variable manually while the CI is still kicked off automatically

### Solution

Re-enable that code.


toros11:
+1

PR #98, opened 2017-01-30
Build archive for vendored projects.

unakatsuo:
We have three binary files to be compiled so that we run ``go build`` command three times. During the time, the files under ``./vendor`` were also re-compiled on each ``go build`` and that causes the longer build time.

I found ``go build -i`` that saves the build archive to ``$GOPATH/pkg/``. It helps two times faster build time in the best case.

Previous

```
real	1m55.531s
user	1m46.112s
sys	0m8.334s
```

``go build -i ./vendor/...`` and then ``go build ./cmd/...``

```
real	1m17.032s
user	1m10.156s
sys	0m6.244s
```

``go build ./cmd/...`` with pre-built archive.

```
real	0m51.680s
user	0m46.868s
sys	0m4.390s
```


unakatsuo:
Applied same change to #106.

PR #99, opened 2017-01-30
Cache cleanup

akirapop:
I have modified the original code to now use the factored out code that was in the rpm_cleanup branch (and now is in master). A dry run gives the correct resulsts -- all cache directories with non-existent or "stale" (no commits within 2 weeks) branches are setup to be removed.


Metallion:
Would it be possible to put it under the garbage collection job before merge? I think it would make the CI layout more clean.

For running them in parallel maybe this can help: https://github.com/jenkinsci/pipeline-examples/blob/master/docs/BEST_PRACTICES.md#parallelism

If it looks like it will take too much time, we could merge already and fix that afterwards.

akirapop:
I have re-structured things now so that the rpm_cleanup and cache_cleanup code are run in parallel as part of one jenkins job (the garbage_collection job). You can take a look at the recent output. It is probably better to look at the "Console Output", as it is clearer to see that the jobs were run in parallel.

Please hold on merging, though: if you look at the "live" run from this morning -- it failed! The reason: there are permission problems.

Metallion:
Code looks good to me. Also looks like the permission problem is solves so is this ok to merge?

akirapop:
Things seem to be in order now -- both the cache cleanup and the rpm cleanup section are able to remove "stale" directories. *BUT:  Note that the problem I mentioned yesterday with some caches now being created by "root". These directories cannot be removed. I've added code that checks to see if the directory owner is root. If so, no rm -rf ...

One note: to change permissions, a "chgrp -R repoci ${RPM_ABSOLUTE}" & "chmod -r g+w ${RPM_ABSOLUTE}" were run.

ci/garbage_collection/cache_cleanups/cache_garbage_collection.sh
```diff
@@ -0,0 +1,29 @@
...
+
+
+    
+. ../garbage_collection_misc.sh
+. ../prune_branches.sh
+
+time_limit=14     ## Days. Set this to give the "deadline". All
+                  ## branches older than this will be removed.
+
+cache_location_dir=/data2/openvdc-ci/branches
```
Metallion: Needs to be changed to `/data` because of the move to dh112. Could be a good idea to read this from the multibox configuration but just hard coding will work for now imo.

ci/garbage_collection/repo_cleanups/rpm_garbage_collection.sh
```diff
@@ -25,7 +25,6 @@ cd ${rpm_base_dir}/master
 nrepos=$(ls -1 . | wc -l)
 if [[ ${nrepos} -lt 2 ]]; then
    echo "Something is wrong. The master directory contains one or less repos. Quitting."
-   exit 1
```
Metallion: This removal makes the above message wrong. :)

Metallion (changes requested):

ci/garbage_collection/cache_cleanups/cache_garbage_collection.sh
```diff
@@ -0,0 +1,29 @@
...
+## main()
+
+
+## Remove all directories whose branch (on git) no longer exists
+## or which has not beenm pushed to within $time_limit days.
+for directory in $(TIME_LIMIT=${time_limit} dirs_to_prune ${cache_location_dir}); do
+
+   group_owner=$(stat -c %G ${cache_location_dir}/${directory})
+   if [[ "${group_owner}" = "root" ]]; then
+       echo "${directory} is owned by root. Cannot remove!"
```
Metallion, line 25: The problem is that currently the acceptance test's cache is always owned by root. This code is ok like this but the multibox code should change the cache images group settings so this script can remove them.

Metallion (changes requested):

Metallion:
There's still the problem where the test cache is always owned by root and thus will never be deleted by this. It would be good to fix that in the same PR. I'll also have a look to see exactly how we should do those settings. The things is that the cache is currently made from a privileged docker container and that's why they're owned by root. A quick hack would probably to do a chown -R or chgrp -R at the end of the acceptance test.

Metallion (approved):

PR #102, opened 2017-01-31
Fix broken acceptance test build

### LINKED ISSUE #95: Multibox seed image broken because of hard coded version
### Problem

The multibox environment currently won't build because the Centos version is hard coded.

```
** DOING STEP: Install zookeeper on zookeeper
++ sudo chroot /home/metallion/work/go/src/github.com/axsh/openvdc/ci/acceptance-test/multibox/10.0.100.10-zookeeper/tmp_root /bin/bash -c 'yum install -y mesosphere-zookeeper'
Loaded plugins: fastestmirror
http://ftp.jaist.ac.jp/pub/Linux/CentOS/7.2.1511/os/x86_64/repodata/repomd.xml: [Errno 14] HTTP Error 404 - Not Found
```

This is a common problem with images generated by [buildbook-rhel7](https://github.com/axsh/buildbook-rhel7).

### Solution

There's two ways we could go about this.

* Change buildbook to use `7` instead of `7.2.1511` in the yum repository.

* Get the seed image from somewhere else.

Metallion:
Fixes https://github.com/axsh/openvdc/issues/95 by copying in fixed yum repositories at the build stage. A better solution would be to fix it in the seed image but this immediately fixes the issue while rebuilding the seed image will take some more time.

### DIFF
--- a/ci/acceptance-test/multibox/10.0.100.10-zookeeper/guestroot/etc/yum.repos.d/CentOS-Base.repo
+++ b/ci/acceptance-test/multibox/10.0.100.10-zookeeper/guestroot/etc/yum.repos.d/CentOS-Base.repo
@@ -1,0 +1,17 @@
+[base]
+name=CentOS-$releasever - Base
+baseurl=http://ftp.jaist.ac.jp/pub/Linux/CentOS/7/os/$basearch/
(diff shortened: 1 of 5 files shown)


toros11:
+1

PR #104, opened 2017-02-01
Fix break after #91 and #94

unakatsuo:
``spf13/cobra`` option parser was enabled in #91 but std ``flags`` option parser was used by the sub-dependency ``golang/glog``. Those two are incompatible for the command line option syntax so that executor failed at the option parsing stage.

Fixes:

- [x] ``-logtostderr`` option for ``glog`` is mandate to dump logs from ``mesos-go`` to mesos-slave. Set the fixed value to the flag and skip option parsing by std ``flag``. 
- [x] ``openvdc-scheduler`` was sending ``-option`` style to executor. Update to ``--option``.

### DIFF
--- a/cmd/openvdc-executor/main.go
+++ b/cmd/openvdc-executor/main.go
@@ -386,12 +386,16 @@ func initConfig() {
 }
 
 func init() {
+	// Initialize golang/glog flags used by mesos-go.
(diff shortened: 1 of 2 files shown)


Metallion (approved):
Looks good to me. :+1:

b0r6:
+1

PR #106, opened 2017-02-01
Port build.sh/build.ps1 to golang

unakatsuo:
It was being hard to maintain two different build scripts for *NIX and Windows. I ported the build steps in those scripts to the single .go file.


unakatsuo:
I wanted to place the file as ``build.go`` to the project root. but ``go`` complains the package name  conflict with ``version.go``.

Which is better?
1. Move ``version.go`` to somewhere then place ``build.go`` to root.
2. No need to change.

b0r6:
I'd rather have build.go in the root instead of version.go

Metallion (approved):
:+1:

PR #112, opened 2017-02-03
Fix #110

### LINKED ISSUE #110: openvdc command fails if home and tmp directories are on other partitions.
### Problem

If the temporary directory and the user's home directory are not on the same partition, `openvdc run centos/7/lxc` will fail with the following error.

```
> ./openvdc run centos/7/lxc
INFO[0000] Updating registry cache from https://raw.githubusercontent.com/axsh/openvdc/master/templates
FATA[0003] Invalid path: centos/7/lxc, rename /tmp/gh-images-reg941395916/openvdc-e77ed15f3b2ba582087afa226ace61a6756f65dd/templates /home/metallion/.openvdc/registry/github.com-axsh-openvdc/master: invalid cross-device link
```

```
> df
...
tmpfs            8099764        52   8099712   1% /tmp
/dev/sda3      469420496 421441484  24110752  95% /home
...
```

The reason is because [os.Rename](https://golang.org/pkg/os/#Rename) doesn't allow moving files between different file systems.

### Temporary workaround

You can set a tmp directory on the same partition using the `TMPDIR` environment variable.

```
> TMPDIR=$HOME/.openvdc/tmp ./openvdc run centos/7/lxc
INFO[0000] Found template: centos/7/lxc
...
```

### Solution

@unakatsuo told me that the `mv` command had to deal with the same problem and solved it by not renaming the file but rather make a copy first and then delete the original. We could try a similar thing in our go code.

unakatsuo:
os.Rename() was used after extracting the resource template archive from github. Rename syscall does not support the moving file over different filesystems.

The resolution is not to call os.Rename(). It is changed to unzip the download archive to ``$HOME/.openvdc/registry`` directly. As part of  this change, unzip() function is improved that it extracts just the specified folder tree. The entier tree was extracted previsously.

Fixes #110.

### DIFF
--- a/registry/github.go
+++ b/registry/github.go
@@ -231,10 +231,15 @@ func unzip(archive, target string) error {
 		return err
 	}
 
+	atime := time.Now()
 	for _, file := range reader.File {
 		path := filepath.Join(target, file.Name)
 		if file.FileInfo().IsDir() {
-			os.MkdirAll(path, file.Mode())
+			err := os.MkdirAll(path, file.Mode())
+			if err != nil {
+				return err
+			}
+			os.Chtimes(path, atime, file.ModTime())
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


Metallion (approved):
Tested it and works. :+1:

PR #118, opened 2017-02-09
Fix rpmbuild failure

Metallion:
Rpmbuild job was failing on all branches including master. I don't know why it succeeded before but read the comment I committed if you want to know what was going on.


b0r6 (approved):
+1

PR #119, opened 2017-02-10
Add diagrams.

unakatsuo:
Using draw.io, generated .svg and .png diagrams below.

- Instance state diagram
- System component overview


Metallion (approved):
This one can skip the CI since it only adds diagrams. :+1: 

Interesting. Maybe I should also switch to draw.io.

PR #125, opened 2017-02-19
Add firewalld service xml

unakatsuo:


pkg/rhel/openvdc.spec
```diff
@@ -85,6 +87,13 @@ This is a 'stub'. An appropriate message must be substituted at some point.
 /opt/axsh/openvdc/bin/openvdc-executor
 %dir /etc/openvdc
 %config(noreplace) /etc/openvdc/executor.toml
+%config(noreplace) /etc/firewalld/services/vdc-executor.xml
+
+%post executor
+%{firewalld_reload}
```
Metallion: This line scares me. Doesn't this mean that the port will be opened automatically after installation? I don't think we can just do that to people's servers. I think it can be helpful to put the services in place but I still think it should be up to the admin to explicitly activate them.

unakatsuo:
https://fedoraproject.org/wiki/PackagingDrafts/ScriptletSnippets/Firewalld addresses ``/usr/lib/firewalld/services/`` for the permanent configuration.

Metallion:
Yes, I read that page before commenting. From what's written there, I get the idea that this configuration will be immediately enabled by the reload macro. I was going to test it but couldn't find a suitable centos7 VM within a reasonable amount of time so I asked here.

It does open a hole in the firewall right away without asking right?

Metallion:
Another way to do it would be to include those services in a separate package. Then if an admin wants to open up a hole in his firewall, s/he could choose to install those packages.

But if we included those with the openvdc metapackage, we have to include clear documentation stating that these ports will be opened.

pkg/rhel/openvdc.spec
unakatsuo: I don't think that the behavior. ``firewall-cmd reload`` is just like ``systemdctl daemon-reload``.

```
$ sudo firewall-cmd --get-default-zone
public
$ sudo firewall-cmd --zone=public --list-all
public (active)
  target: default
  icmp-block-inversion: no
  interfaces: eth0 eth1
  sources:
  services: dhcpv6-client ssh
  ports:
  protocols:
  masquerade: no
  forward-ports:
  sourceports:
  icmp-blocks:
  rich rules:
$ sudo vi /usr/lib/firewalld/services/vdc-executor.xml
$ sudo firewall-cmd --get-services | grep vdc-executor
$ sudo firewall-cmd --reload
success
$ sudo firewall-cmd --get-services | grep vdc-executor
RH-Satellite-6 amanda-client amanda-k5-client bacula bacula-client ceph ceph-mon dhcp dhcpv6 dhcpv6-client dns docker-registry dropbox-lansync freeipa-ldap freeipa-ldaps freeipa-replication ftp high-availability http https imap imaps ipp ipp-client ipsec iscsi-target kadmin kerberos kpasswd ldap ldaps libvirt libvirt-tls mdns mosh mountd ms-wbt mysql nfs ntp openvpn pmcd pmproxy pmwebapi pmwebapis pop3 pop3s postgresql privoxy proxy-dhcp ptp pulseaudio puppetmaster radius rpc-bind rsyncd samba samba-client sane smtp smtps snmp snmptrap squid ssh synergy syslog syslog-tls telnet tftp tftp-client tinc tor-socks transmission-client vdc-executor vdsm vnc-server wbem-https xmpp-bosh xmpp-client xmpp-local xmpp-server
$ sudo firewall-cmd --zone=public --list-all
public (active)
  target: default
  icmp-block-inversion: no
  interfaces: eth0 eth1
  sources:
  services: dhcpv6-client ssh
  ports:
  protocols:
  masquerade: no
  forward-ports:
  sourceports:
  icmp-blocks:
  rich rules:

```



```
$ rpm --showrc | grep firewall
-14: firewalld_reload
test -f %{_bindir}/firewall-cmd && firewall-cmd --reload --quiet || :
```

Metallion:
As far as I understand, this will still open up a hole in the firewall without asking. The difference is just that it won't happen right away. Am I wrong?

Metallion:
I'm closing this one due to lack of activity. Making services for firewalld is a great idea but it should be done in such a way that it doesn't apply them without asking.

I would suggest making them into separate packages `openvdc-scheduler-firewalld` and `openvdc-executor-firewalld` that can be installed optionally.

PR #127, opened 2017-02-20
Scheduler update

b0r6:
This makes it possible for the user to configure a few more things about the framework:
* ID - for example "My framework" instead of random ones generated by mesos like: 4f6dd761-f65b-4830-9e55-3c319341593e-0003
* Name
* Failover timeout - If a framework (scheduler) crashes mesos master kills all the associated executors and tasks by default. But if a failover timeout is set, the master waits a certain amount of time to the scheduler to come back up before killing anything.
* Executor-path

So now, even if the scheduler crashes, the containers will keep on running. Also since the frameworkID will remain the same once the scheduler is back up, commands that uses frameworkmsg will still work.

The scheduler now also assigns resources to executors depending on the resource template file.


unakatsuo (approved):

pkg/conf/scheduler.toml
```diff
@@ -7,3 +7,9 @@
 
 [api]
 # endpoint = "localhost:5000"
+
+[scheduler]
+# id = "scheduler_1"
+# name = "openvdc"
+# failover-timeout = 604800 #1 week
+# executor-path = "openvdc-executor"
```
Metallion, line 15: Very good to see these added to the default config file as they should. Something that is often forgotten. :+1:

Metallion (approved):

PR #128, opened 2017-02-28
Chown fix in acceptance test docker script

Metallion:
The acceptance test needs to make jenkins the owner of its cache images so the garbage collection job can delete them later without using sudo.

This was done in the main script after `docker run` which caused it to not be executed after if the tests (and thus the `docker run` command) failed. I moved it to the cleanup function so it is always executed.

While I was at it I also changed it so it uses whatever the current user is instead of jenkins.


akirapop (commented):
Looks clean to me.

akirapop (approved):

Metallion:
Thanks for the review. The CI is currently failing on all branches because the jaist centos mirror that we're depending on is having trouble. Gonna check if we can't depend on multiple mirrors.

PR #129, opened 2017-03-02
Make console command work for non tty mode

unakatsuo:
``openvdc console`` currently supports interactive mode but the no-interactive mode is not working correctly.

The goal is to accept following command line syntax:

```
echo ls | openvdc console i-xxxxxxxx
cat myscript.sh | openvdc console i-xxxxxxxx
openvdc console i-xxxxxxxx -- ls
```

- [x] Split ``Console`` interface to tty (PtyConsole) and non-tty (Console) version
- [x] Handle ``env`` ssh request
- [x] unix: Handle ``SIGWINCH`` and reflect client terminal window size change
- [x] Handle ``exec`` ssh request
- [x] ``echo ls | openvdc console`` continues to wait next input
- [x] Tell local ``$TERM`` value and make it effective on remote.
- [x] Return ``exit-status`` SSH response

Windows console support will be continued in another PR.
-  windows: Handle ``ReadConsoleInput()`` and reflect client terminal window size change
    - https://msdn.microsoft.com/en-us/library/windows/desktop/ms685035(v=vs.85).aspx
-  windows: ``Write-Out ls | openvdc console`` sends with ``\r`` so the command string is not recognized on remote.
- windows:  ``Write-Out ls | openvdc console`` crashes the executor
-  windows: Cursor keys does not work in the interactive shell. Other special keys as well probably.
- Return ``exit-signal`` SSH response


unakatsuo:
I realized that it takes more time to support windows console properly. It should be completed by another PR. TODO list will be re-organized to be satisfy with UNIX tty.

Metallion:
Needs to resolve merge conflicts.

b0r6:
Looks good to me. Also gave it a try and tested a bunch of things, works great.

b0r6 (approved):

PR #130, opened 2017-03-03
Update build.go

unakatsuo:
- Set ``ignore`` build tag
- Guess some  bin paths  from go environemt variable and append to $PATH

### DIFF
--- a/build.go
+++ b/build.go
@@ -103,6 +103,13 @@ func determineGHRef() string {
 }
 
 func main() {
+	// Apppend $GOBIN and $GOPATH/bin to $PATH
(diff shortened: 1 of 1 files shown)


Metallion:
Why are you adding the bin directory to PATH? Is it for the call to govendor [here](https://github.com/axsh/openvdc/blob/bb1ca8bbd9544f70c3f0b9689feb409d5a41555f/build.go#L179)?

unakatsuo:
For ``govendor`` and ``go-bindata`` they are installed ``$GOPATH/bin``. Developers still have to maintain the location for ``protoc`` binary.

Metallion (approved):

PR #132, opened 2017-03-08
Remove Resource model and API

unakatsuo:
``Resource`` model was introduced but it was not well designed for real use case. Generally we have been just using ``openvdc run <template.json>`` or ``openvdc create <template.json>`` to get new instance then run ``openvdc destroy <ID>`` to finalize the item. I think this workflow would not be changed if we extend the targets to other datacenter resources in near future.

So we don't need the Resource model which was to associating between template and instance models. The command line/gRPC API operations required extra calls to deal with the model.  ``openvdc run`` hide those calls so that not much people weren't aware of.
 
Comparison for minimal primitive calls.

Previous:

```
% openvdc register <template.json>
r-xxxx
% openvdc create r-xxxx  # i-xxxx was created based on r-xxxx definition
i-xxxx
% openvdc destroy i-xxxx
% openvdc unregister r-xxxx
```

After this change:

```
% openvdc create <template.json>
i-xxxx
% openvdc destroy i-xxxx
```

- [x] Remove definition from ``model.proto`` and ``v1.proto``
- [x] Remove ``openvdc register``, ``openvdc unregister``
- [x] Remove ``Resource`` gRPC API endpoint


b0r6 (approved):

Metallion:
Waiting for review from @b0r6 on this one. :)

b0r6:
+1

Metallion:
Was going to merge but waiting for the CI result because new commits were made.

PR #133, opened 2017-03-10
Fix openvdc template validate

unakatsuo:
JSON template validation function ``registry.ValidateTemplate()`` exists but it was called nowhere.


ci/acceptance-test/tests/cmd_template_test.go
```diff
@@ -0,0 +1,27 @@
...
+func TestCmdTemplateValidate(t *testing.T) {
+	RunCmdAndReportFail(t, "openvdc", "template", "validate", "centos/7/lxc")
+	RunCmdAndReportFail(t, "openvdc", "template", "validate", "/var/tmp/fixtures/lxc.json")
+	RunCmdAndReportFail(t, "openvdc", "template", "validate", "https://raw.githubusercontent.com/axsh/openvdc/master/templates/centos/7/lxc.json")
+}
+
+func TestCmdTemplateShow(t *testing.T) {
+	RunCmdAndReportFail(t, "openvdc", "template", "show", "centos/7/lxc")
+	RunCmdAndReportFail(t, "openvdc", "template", "show", "/var/tmp/fixtures/lxc.json")
+	RunCmdAndReportFail(t, "openvdc", "template", "show", "https://raw.githubusercontent.com/axsh/openvdc/master/templates/centos/7/lxc.json")
```
Metallion: Here I don't think it is enough to just check the exit status. What if the commands don't show anything or show something completely wrong and still exit with 0?

ci/acceptance-test/tests/cmd_template_test.go
```diff
@@ -0,0 +1,27 @@
...
+func init() {
+	if err := RestoreAsset("/var/tmp", "fixtures/lxc.json"); err != nil {
+		panic(err)
+	}
+}
+
+func TestCmdTemplateValidate(t *testing.T) {
+	RunCmdAndReportFail(t, "openvdc", "template", "validate", "centos/7/lxc")
+	RunCmdAndReportFail(t, "openvdc", "template", "validate", "/var/tmp/fixtures/lxc.json")
+	RunCmdAndReportFail(t, "openvdc", "template", "validate", "https://raw.githubusercontent.com/axsh/openvdc/master/templates/centos/7/lxc.json")
```
Metallion: Is it enough to only check the exit status here? If it is, maybe also include some negative tests to confirm that the validation command will indeed fail when called on an invalid template.

Metallion (changes requested):

ci/citest/acceptance-test/tests/cmd_template_test.go
```diff
@@ -0,0 +1,27 @@
...
+func init() {
+	if err := RestoreAsset("/var/tmp", "fixtures/lxc.json"); err != nil {
+		panic(err)
+	}
+}
+
+func TestCmdTemplateValidate(t *testing.T) {
+	RunCmdAndReportFail(t, "openvdc", "template", "validate", "centos/7/lxc")
+	RunCmdAndReportFail(t, "openvdc", "template", "validate", "/var/tmp/fixtures/lxc.json")
+	RunCmdAndReportFail(t, "openvdc", "template", "validate", "https://raw.githubusercontent.com/axsh/openvdc/master/templates/centos/7/lxc.json")
```
Metallion: Is it enough to only check the exit status here? If it is, maybe also include some negative tests to confirm that the validation command will indeed fail when called on an invalid template.

ci/citest/acceptance-test/tests/cmd_template_test.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+func TestCmdTemplateShow(t *testing.T) {
+	RunCmdAndReportFail(t, "openvdc", "template", "show", "centos/7/lxc")
+	RunCmdAndReportFail(t, "openvdc", "template", "show", "/var/tmp/fixtures/lxc.json")
+	RunCmdAndReportFail(t, "openvdc", "template", "show", "https://raw.githubusercontent.com/axsh/openvdc/master/templates/centos/7/lxc.json")
```
Metallion: Here I don't think it is enough to just check the exit status. What if the commands don't show anything or show something completely wrong and still exit with 0?

Metallion (changes requested):
Repeated my comments because github had marked them as outdated because of a commit that moved the files but didn't solve the problems I commented about.

Metallion (approved):

PR #145, opened 2017-03-25
Use fixed port number for executor sshd

unakatsuo:
The listen port number for the embedded sshd was chosen randomly by default. That decision just made the code for initialization complex. So it listens TCP ``29876`` port by default. Users still can apply arbitrary listen port from ``console.ssh.listen`` configuration parameter.


Metallion (approved):
👍

PR #146, opened 2017-03-28
executor configuration sub RPM package

unakatsuo:
``openvdc-executor.rpm`` has been organized to each hypervisor driver configuration package.

- ``openvdc-executor.rpm`` contains binary and shared folders
- ``openvdc-executor-lxc.rpm`` has dependencies to lxc-* and openvdc-executor rpm
- ``openvdc-executor-null.rpm`` has dependency to openvdc-executor rpm


Metallion (approved):
👍

PR #147, opened 2017-03-29
Fix issue translation protobuf field name to JSON key.

unakatsuo:
``protoc`` compiler seems to change the rule of JSON key naming convention in recent releases.

Going forward, we need to add ``[json_name="..."]`` to each protobuf message field if it includes "_" otherwise the camel case rule is applied to JSON keys.

https://github.com/google/protobuf/blob/master/src/google/protobuf/descriptor.proto#L196-L200

```
message Instance {
   string id = 1;
   string slave_id = 2 [json_name="slave_id"];
```

Also ``build.go`` was changed to enforce developers to use the exact same version of protoc.


Metallion (approved):
Merging as the combination of this and 2 more PRs will make master green.

PR #154, opened 2017-04-10
Inherit mesos-slave attributes

unakatsuo:
``/etc/mesos-slave/attributes`` and ``/etc/openvdc/executor.toml`` are having duplicate configuration items.

- hypervisor
- openvdc-node-id 

This PR removes duplications from executor.toml and retrieve the attributes above from mesos-slave.


Metallion:
Why keep a png of `docs/diagrams/crash-recovery.png`  when we have a svg of the same illustration? I think using png in git should be avoided because it's inefficient. Every time the illustration changes, a new copy of the whole file (which is rather large) will be added in git. An svg can just place the lines that changed.

Other than that i'm waiting for a review from @b0r6 and for the merge conflict to be resolved. ^_^

unakatsuo:
Someone was discussing that .svg is not rendered in GH Markdown. It supports raster image format only.

https://github.com/axsh/openvdc/blob/acceptance-test-ovs-executor/ci/citest/acceptance-test/README.md

Metallion:
I know about it not being rendered in GH markdown but does it need to be in this case? i imagine the docs folder will be used to generate a static site, probably using mkdocs like we do for Wakame-vdc and OpenVNet.

Metallion:
Still waiting for a review from @b0r6 and either the removal of that png or an answer to my comment about the docs folder. :)

b0r6:
+1

Though it contains the crash-recovery branch, right? Which hasn't been merged yet due to that missing test that I plan to write soon.

Metallion:
Oh right... I actually hadn't realized that yet. 😮  Thanks. :) In that case we still need to get that tested first, indeed.

Issue #156, opened 2017-04-17
Instance stuck on `STARTING` if no internet access on executor


Metallion:
### Problem

When you start an instance but the executor instance does not have access to the internet, the instance will never come up and its database entry will be stuck in `STARTING` state forever.

### Reproduction

* Install `openvdc-cli` on machine A.
* Install `openvdc-executor-lxc` on machine B.
* Set it up so machine A and machine B have network access to each other but machine B does not have internet access. For example remove its default gateway from the routing table.
* On machine A: 

```
A $ openvdc run centos/7/lxc
INFO[0000] Updating registry cache from https://raw.githubusercontent.com/axsh/openvdc/master/templates
INFO[0002] Found template: centos/7/lxc
i-0000000000
```
* On machine B:

```
B $ lxc-info -n i-0000000000
i-0000000000 doesn't exist
```

* On machine A:

```
A $ openvdc show i-0000000000
{
  "iD": "i-0000000000",
  "instance": {
    "id": "i-0000000000",
    "slaveId": "91fb613c-2abf-477a-b41c-41b37e1845b9-S0",
    "lastState": {
      "state": "STARTING",
      "createdAt": "2017-04-17T06:48:54.339487647Z"
    },
    "createdAt": "2017-04-17T06:48:48.379337475Z",
    "template": {
      "templateUri": "https://raw.githubusercontent.com/axsh/openvdc/master/centos/7/lxc.json",
      "lxc": {
        "lxcTemplate": {
          "template": "download",
          "distro": "centos",
          "release": "7"
        }
      },
      "createdAt": "2017-04-17T06:48:48.379337475Z"
    }
  }
}
```

Instance `i-0000000000` is now stuck in this state forever.

### Suggested solution

I would suggest a `FAILED` state which is identical to `TERMINATED` except it happened because of an error and not because a user called `openvdc destroy`. Something like this:

```
 $ openvdc show i-0000000000
{
  "iD": "i-0000000000",
  "instance": {
    ...
    "lastState": {
      "state": "FAILED",
      "Reason: Failed to download lxc template: <whatever error message golang gave us>"
      ...
    },
  ...
}
```

unakatsuo:
agree with the solution. will start the new PR for this issue.

unakatsuo:
![instance-state](https://cloud.githubusercontent.com/assets/95236/25085042/05de3972-239b-11e7-9b08-5e368f7c6234.png)

Metallion:
Confirmed fixed.

PR #161, opened 2017-05-09
Fall back to "FAILED" state

unakatsuo:
#156 

- [x] Defined "FAILED" state
- [x] executor handles the new state in case of error.
- [x] Extend null driver to be able to simulate crash at specific stage.
- [x] Write ci/tests to crash the null instance and check if its state value becomes "FAILED".


docs/diagrams/instance-state.svg
```diff
@@ -1,2 +1,2 @@
...
...form="translate(269.5,732.5)"><switch><foreignObject style="overflow:visible;" pointer-events="all" width="41" height="16" requiredFeatures="http://www.w3.org/TR/SVG11/feature#Extensibility"><div xmlns="http://www.w3.org/1999/xhtml" style="display: inline-block; font-size: 15px; font-family: Helvetica; color: rgb(0, 0, 0); line-height: 1.2; vertical-align: top; width: 42px; white-space: nowrap; word-wrap: normal; text-align: center;"><div xmlns="http://www.w3.org/1999/xhtml" style="display:inline-block;text-align:inherit;text-decoration:inherit;">Failed</div></div></foreignObject><text x="21" y="16" fill="#000000" text-anchor="middle" font-size="15px" font-family="Helvetica">Failed</text></switch></g><ellipse cx="87" cy="816" rx="11" ry="11" fill="#000000" stroke="#ff0000" pointer-events="none"/><ellipse cx="87" cy="816" rx="15" ry="15" fill="none" stroke="#ff0000" pointer-events="none"/><ellipse cx="290" cy="816" rx="11" ry="11" fill="#000000" stroke="#ff0000" pointer-events="none"/><ellipse cx="290" cy="816" rx="15" ry="15" fill="none" stroke="#ff0000" pointer-events="none"/><ellipse cx="87" cy="36" rx="11" ry="11" fill="#000000" stroke="#ff0000" pointer-events="none"/></g></svg>
```
Metallion: You should use xmllint to turn this in a multiline xml file. Everything on one line is great for browsers but not for git. It would cause every change to the file to replace the entire file.

`xmllint --format old_file.svg > new_file.svg`

Metallion (changes requested):
* The openvdc-acceptance-test binary file should not be comitted. I assume this was accidental but anyway it needs to be removed. :)

* As I mentioned in another PR I think we shouldn't keep .png files in the docs folder. Svg is enough. The security issue you mentioned in the other PR only applies when we want to render them directly in markdown files on github.

unakatsuo:
the binary was committed wrongly at b3837d84ec. I removed it and rebased the changes after that commit.

b0r6 (approved):

Metallion:
Thanks for removing the binary. What about my comment about the .png file? If you disagree, let's discuss but please reply. :)

unakatsuo:
That topic is out of this thread. But I think attaching .png gives online preview on github. It helps reviewing for what changed.

Metallion:
> That topic is out of this thread. 

No, there is a png file being added in this pull request so the topic of whether png files should be added does belong in this thread.

> But I think attaching .png gives online preview on github. It helps reviewing for what changed.

Try clicking on the "display the rich diff" button of the .svg file. :-) It's the button with a file shaped icon next to the <> button.

unakatsuo:
I didn't know about the botton. it's a good time to know. My concern has gone away.

Metallion (approved):

PR #166, opened 2017-05-22
Update CI seed image

Metallion:
The CI was failing because the CentOS base repository depends on jaist servers which are down at the moment. I removed those repos and updated the seed image.

The new seed image:

* Runs Centos 7.3 instead of 7.2.
* Has a CentOS base repository with a mirrorlist configured instead of a baseurl. If one mirror is down, it will just use another one.


unakatsuo:
Want to add ``fastestmirror.conf`` to limit mirror sites in JP.

/etc/yum/pluginconf.d/fastestmirror.conf

```
[main]
enabled=1
verbose=0
always_print_best_host = true
socket_timeout=3
#  Relative paths are relative to the cachedir (and so works for users as well
# as root).
hostfilepath=timedhosts.txt
maxhostfileage=10
maxthreads=15
#exclude=.gov, facebook
#include_only=.nl,.de,.uk,.ie
include_only=.jp
```

Metallion:
Ok, I will add that.

unakatsuo (approved):

Issue #169, opened 2017-05-23
OVS networking scripts missing


Metallion:
### Problem

When using Open vSwitch starting an instance crashes with the following error.

```
2017-05-23 07:23:49 [ERROR] openvdc-executor/main.go:158 Failed CreateInstance Error: Failed to parse script template: /etc/openvdc/scripts/ovs-up.sh.tmpl: open /etc/openvdc/scripts/ovs-up.sh.tmpl: no such file or directory hypervisor=lxc instance_id=i-0000000007 state=STARTING
ErrorStackTrace:
github.com/axsh/openvdc/hypervisor/lxc.(*LXCHypervisorDriver).renderUpDownScript
        /var/tmp/go/src/github.com/axsh/openvdc/hypervisor/lxc/lxc.go:230
github.com/axsh/openvdc/hypervisor/lxc.(*LXCHypervisorDriver).CreateInstance
        /var/tmp/go/src/github.com/axsh/openvdc/hypervisor/lxc/lxc.go:301
```

It looks like the `/etc/openvdc/scripts` directory is missing entirely.

```
# ls /etc/openvdc/
executor.toml  scheduler.toml  scheduler.toml.rpmnew
```

### Solution

Put the required scripts for Open vSwitch back in place.

PR #173, opened 2017-05-29
Added executor-lxc to meta package

b0r6:
#169 

Meta package didn't include executor-lxc, so the scripts didn't get installed.


Metallion (approved):
Thanks for the fix.

Issue #170, opened 2017-05-23
Write tests with Open vSwitch


Metallion:
### Problem

We are supporting Open vSwitch but we do not have any tests for OpenVDC with Open vSwitch yet.

### Dependencies

We currently have an Open vSwitch host on the CI but we have no means of making sure it gets used in specific tests yet. This feature is being developed in https://github.com/axsh/openvdc/pull/159.

### Solution

Finish https://github.com/axsh/openvdc/pull/159 first and then write some tests.

PR #175, opened 2017-06-01
Ci acceptance test ovs

toros11:
Solves #170 
Possibly we should cover more scenarios where the openvswitch node get's used


unakatsuo (approved):

Metallion (approved):
:+1:

PR #168, opened 2017-05-22
Local image

b0r6:
#162

PR #176, opened 2017-06-03
Fix for failing local_imate_test.go occasinally.

unakatsuo:
``ci/citest/tests/local_image_test.go`` runs the test instance without ``node_groups: ["linuxbr"]`` attribute. so that the later ssh checking is failing sometime.

https://github.com/axsh/openvdc/pull/168#issuecomment-305739019


Metallion:
I am going to run the CI a few more times to see if it is indeed fixed.

Metallion (approved):
Ran the CI 4 times and succeeded every time. I think we're good.

PR #177, opened 2017-06-05
Build installs protoc

ChrisSalisbury:
Update the build script to automatically install the protoc binary version that we recommend. Thinking about it now, I realize that I hardcoded the version rather than use the global that is hardcoded -- I'll change that in a minute so that changing the hardcoded global (or making that variable not hardcoded as is hopefully planned) will also change the binary that gets installed.


Metallion:
Looks like the CI failure might be this: https://github.com/axsh/openvdc/pull/176

I'm currently running the CI on there a couple of times to check if it indeed succeeds every time. If it does, I will merge it and then run the CI here again. Since the CI always merges in the latest master locally, this should then succeed as well.

Metallion:
#176 was merged. Rerunning the CI on this.

Edit: As expected the CI ran successfully this time. ^_^

README.md
```diff
@@ -5,6 +5,7 @@ Extendable Tiny Datacenter Hypervisor on top of Mesos architecture. Wakame-vdc v
 ## Build
 
 Ensure ``$GOPATH`` is set. ``$PATH`` needs to have ``$GOPATH/bin``.
+The developer package of lxc is also required. (lxc-devel from epel-release on redhat systems, or lxc-dev on debian systems) package is also required.
```
Metallion: I think an extra newline between GOPATH and lxc-dev would make it easier to read but not gonna hold back the PR for that.

build.go
```diff
@@ -68,6 +71,139 @@ func if_not_exists(cmd string, cb func()) {
...
+
+    // Closure to address file descriptors issue with all the deferred .Close() methods
+    extractAndWriteFile := func(f *zip.File) error {
+       rc, err := f.Open()
+        if err != nil {
+            return err
+        }
+        defer func() {
+            if err := rc.Close(); err != nil {
+                panic(err)
```
Metallion, line 99: I was under the impression that panic was something to be used sparingly in go. It also seems this code doesn't intend to recover from this panic further up the stack. Why choose it instead of printing the error and exiting?

build.go
```diff
@@ -68,6 +71,139 @@ func if_not_exists(cmd string, cb func()) {
...
+	return strings.Join(args, separator)
+}
+
+func installProtoc() {
+		if_not_exists("protoc", func() { // since we require a specific version anyway, wouldn't it be better to include this binary in the initial install filesystem?
+			const GOARCH string = runtime.GOARCH
+			const GOOS string = runtime.GOOS
+			if GOOS == "linux" && GOARCH == "amd64" { // check os and arch. If 64-bit linux, download binary. Other cases should be added as we start supporting other oses/architectures.
+				protocVersionNumber := strings.Split(ProtocVersion, " ")[1]
+				filename := downloadFromUrl(join("", "https://github.com/google/protobuf/releases/download/v", protocVersionNumber, "/protoc-", protocVersionNumber, "-linux-x86_64.zip"))
```
Metallion, line 179: Why is it necessary to download this explicitly instead of using `govendor` like the rest of the dependencies?

build.go
```diff
@@ -136,7 +272,8 @@ Environment Variables:
 	}
 	BUILDDATE := time.Now().UTC().Format(time.RFC3339)
 
-	GOVERSION := cmd("go", "version")
+	GOVERSION := runtime.Version() //cmd("go", "version")
```
Metallion: No need to keep comments of old lines imo. Git keeps those in the commit history for us. :)

README.md
ChrisSalisbury: I'm happy to add a newline...give me a minute.

build.go
ChrisSalisbury: You're probably right -- I left this how it was when I found it. Maybe we have to go back through blame to ask the right person?

build.go
ChrisSalisbury: I agree completely. I commented about it in the code at the top. I think that is definitely the better solution. We should have all the binaries for the different oses and architectures and just choose the right one when installing. But I don't know how our vendoring system works. I would just include it in the package directly...but then again, I would just have the whole thing in a package to begin with rather than run a build script..

build.go
ChrisSalisbury: agreed. This is again, not mine..

build.go
ChrisSalisbury: err. wait. That one was mine. Let me take a closer look.

ChrisSalisbury: ok, right, I remember now. This is a deferred file close with potentially more than one. Panic lets all deferred operations finish before exiting and deferred functions execute right before close. So this should be a safe panic and has a reason.

build.go
ChrisSalisbury: man, I should look at these better. This one was mine, too. Just forgot to delete it. Sorry.

build.go
Metallion: I see. Thanks for the explanation. :)

build.go
Metallion: The way dependencies are currently handled is by putting them in this file: https://github.com/axsh/openvdc/blob/master/vendor/vendor.json

The `build.go` script then uses govendor to download the exact revisions described in them. Interestingly protobuf seems to be defined in there though. I wonder why it's referencing specific files from the protobuf repo though. Perhaps @unakatsuo knows more?

ChrisSalisbury: In this case, I don't see any real difference here. If google decides to change their legacy repo of this binary to something that no longer works, and does it as a commit, than the govendor method would be superior, but the odds of that seem very similar to the odds of them deleting the legacy binaries all together. So it would reduce the currently low risk to a lower risk. Seems like it would make more sense to just eliminate the risk. But having said that, if I get some time to do it, I will try to remember to change the download method to govendor...but it's pretty low priority in my mind.

unakatsuo:
I think installation for tool chains should be handled out of the build script. The build script just needs to check if each tool exist.

I wrote similar steps before.

https://github.com/axsh/openvdc/blob/master/ci/devbox/devbox-centos7.json

unakatsuo:
``unzip`` function seems to be copied from stackoverflow. you should add the credit to the comment.

ChrisSalisbury:
If you don't think that we should be installing requirements into our filesystem that don't exist, then this branch should just be deleted and the PR rejected. Easy enough. I disagree (I wouldn't have written this if I didn't of course), but feel free to require whatever you like of your users. I will try to remember.

Regarding the unzip function, I did definitely use references (it was more than a few, I don't recall if stackoverflow was one of them or not) and did not think to cite them. I suppose I will try to keep that in mind in the future as well.

Metallion:
Closing as it was decided not to install OS specific dependencies in the build script.

PR #178, opened 2017-06-05
Refactor download images to bashsteps

toros11:
Refactor to make downloading of images use bashsteps to perform the tasks.
The images also gets placed in the guestroot to avoid use of scp.

### DIFF
--- a/ci/citest/acceptance-test/multibox/ind-steps/step-vdc-images/init.sh
+++ b/ci/citest/acceptance-test/multibox/ind-steps/step-vdc-images/init.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+for img in "${IMAGES[@]}" ; do
+    item=$(grep "$img" <<< "$imgIndex")
+    IFS=';' read -a parts <<< "$item"
+    IFS=';' eval 'folders=($item)'
+    imgRemoteDir="${parts[-1]}"
+
+    prepare_container_image "${img}" "${folders[0]}/${folders[1]}/${folders[2]}" "$imgRemoteDir"
+    add_container_image "${img}" "${folders[0]}/${folders[1]}/${folders[2]}"
+done
+
+(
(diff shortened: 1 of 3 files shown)


Metallion:
Looks like the failures here were also the problem fixed in https://github.com/axsh/openvdc/pull/176. Since that is merged now, the CI should now succeed. Running it again here.

ci/citest/acceptance-test/multibox/ind-steps/step-vdc-images/init.sh
```diff
@@ -0,0 +1,21 @@
...
+    IFS=';' eval 'folders=($item)'
+    imgRemoteDir="${parts[-1]}"
+
+    prepare_container_image "${img}" "${folders[0]}/${folders[1]}/${folders[2]}" "$imgRemoteDir"
+    add_container_image "${img}" "${folders[0]}/${folders[1]}/${folders[2]}"
+done
+
+(
+    # resync again after images have been downloaded
+    $starting_step "Synching guestroot for ${vm_name}"
```
Metallion, line 15: It's a little inefficient to sync twice. I understand how you're feeling though. I've also ran into problems where I wanted to do something before rsync but wasn't able to. Therefore It might be better to rethink the way our steps are set up. I'm thinking something like this.

```
pre_rsync.sh
*call rsync*
post_rsync.sh
pre_boot.sh
*boot vm*
post_boot.sh
```

Doesn't necessarily need to be in this pull request since it's a bigger more all encompassing design change but I'm also a little afraid of postponing such refactors and letting code like this build up. What's your opinion?

toros11: Sounds like a reasonable change, rsync itself might as well be a stage then everything else can be kept as it is.

Metallion (approved):

PR #181, opened 2017-06-16
Adding client authentication with public key fuction.

itouri:
```
$ ./openvdc console -h
Connect to an instance.

Usage:
  openvdc console [Instance ID] [options] [--] [commands] [flags]

Examples:

% openvdc console i-000000001
root@i-00000001#

% echo ls | openvdc console i-000000001
% cat test.sh | openvdc console i-000000001


Flags:
  -i, --identity-file string   Selects a file from which the identity (private key) for public key authentication is read
      --show                   Show console information
```


ci/citest/acceptance-test/tests/cmd_console_test.go
```diff
@@ -23,3 +23,14 @@ func TestCmdConsole_ShowOption(t *testing.T) {
...
+
+func TestCmdConsole_Authentication(t *testing.T) {
+	stdout, _ := RunCmdAndReportFail(t, "openvdc", "run", "centos/7/lxc", `{"authentication_type":"none"}`)
+	instance_id := strings.TrimSpace(stdout.String())
+
+	WaitInstance(t, 5*time.Minute, instance_id, "RUNNING", []string{"QUEUED", "STARTING"})
+
+
+	RunCmdWithTimeoutAndReportFail(t, 10, 5, "openvdc", "destroy", instance_id)
+	WaitInstance(t, 5*time.Minute, instance_id, "TERMINATED", nil)
```
Metallion: How is this a test for console authentication? It just starts an instance and then terminates it.

Metallion (changes requested):

handlers/vm/base.go
```diff
@@ -12,6 +23,84 @@ var SupportedAPICalls = []string{
...
+	// Check that the key is in OpenSSH format.
+	keyNames := []string{"ssh-rsa", "ssh-dss", "ecdsa-sha2-nistp256", "ssh-ed25519"}
+	firstStr := strings.Fields(keyStr)
+	for _, name := range keyNames {
+		if firstStr[0] == name {
+			return true
+		}
+	}
+
+	// Check that the key is in SECSH format.
```
unakatsuo: We don't need to support SECSH format at this version.

handlers/vm/base.go
```diff
@@ -12,6 +23,84 @@ var SupportedAPICalls = []string{
...
+
+func validatePublicKey(key []byte) bool {
+	// Check that the key is in RFC4253 binary format.
+	_, err := ssh.ParsePublicKey(key)
+	if err == nil {
+		return true
+	}
+
+	keyStr := string(key[:])
+	// Check that the key is in OpenSSH format.
```
unakatsuo: Why do you check same thing that happens in ``ssh.ParsePublicKey()``?

https://github.com/golang/crypto/blob/eb71ad9bd329b5ac0fd0148dd99bd62e8be8e035/ssh/keys.go#L246-L258

handlers/vm/base.go
```diff
@@ -12,6 +23,84 @@ var SupportedAPICalls = []string{
...
+
+func (*Base) ValidatePublicKey(h handlers.ResourceHandler, authType model.AuthenticationType, sshPubKey string) error {
+	switch authType {
+	case model.AuthenticationType_NONE:
+	case model.AuthenticationType_PUB_KEY:
+		if sshPubKey == "" {
+			return handlers.ErrInvalidTemplate(h, "ssh_public_key is not set")
+		}
+
+		isValidate := validatePublicKey([]byte(sshPubKey))
```
unakatsuo: Just call ``ssh.ParsePublicKey()`` then return the error from the function. This function squash the error details from ``ssh.ParsePublicKey()``.

unakatsuo (changes requested):

handlers/vm/base.go
itouri: ssh.ParsePublicKey() can parse only RFC4253 format key. Therefore if we want to parse other type public key then we must parse by own.

handlers/vm/base.go
itouri: Changed that part into a comment.

handlers/vm/base.go
itouri: Changed the return type of validatePublicKey([]byte) to error.

handlers/vm/base.go
unakatsuo: I see.

cmd/openvdc-executor/sshd.go
```diff
@@ -22,9 +23,52 @@ type SSHServer struct {
...
+		return nil, err
+	}
+	instResource := inst.ResourceTemplate().(model.InstanceResource)
+	return instResource, nil
+}
+
 func NewSSHServer(provider hypervisor.HypervisorProvider, ctx context.Context) *SSHServer {
 	config := &ssh.ServerConfig{
-		NoClientAuth: true,
+		PasswordCallback: func(conn ssh.ConnMetadata, pass []byte) (*ssh.Permissions, error) {
```
unakatsuo, line 41: Why is the password authentication enabled? Just ``NONE`` and ``PUB_KEY`` are listed in ``model.proto``.

cmd/openvdc/cmd/console.go
```diff
@@ -46,6 +51,10 @@ var consoleCmd = &cobra.Command{
 		}
 
 		var res *api.ConsoleReply
+
+		info, _ := cmd.Flags().GetBool("show")
+		fmt.Printf("show is %v", info)
```
unakatsuo: Remove printf debug.

cmd/openvdc/cmd/console.go
```diff
@@ -56,7 +65,8 @@ var consoleCmd = &cobra.Command{
 			log.WithError(err).Fatal("Failed request to Instance.Console API")
 		}
 
-		info, err := cmd.Flags().GetBool("show")
+		// info, err := cmd.Flags().GetBool("show")
+		// fmt.Printf("show is %v", info)
```
unakatsuo: Remove printf debug.

handlers/vm/base.go
```diff
@@ -12,6 +23,84 @@ var SupportedAPICalls = []string{
...
+	}
+
+	// // Check that the key is in SECSH format.
+	// keyNames = []string{"SSH2 ", "RSA", ""}
+	// for _, name := range keyNames {
+	// 	if strings.Contains(keyStr, "---- BEGIN "+name+"PUBLIC KEY ----") &&
+	// 		strings.Contains(keyStr, "---- END "+name+"PUBLIC KEY ----") {
+	// 		return nil
+	// 	}
+	// }
```
unakatsuo: Remove commented old code.

unakatsuo (changes requested):

cmd/openvdc/cmd/console.go
itouri: I removed printf debug.

cmd/openvdc/cmd/console.go
itouri: I removed printf debug.

handlers/vm/base.go
itouri: I removed old code.

cmd/openvdc-executor/sshd.go
itouri: Library of SSH can't call PublickeyCallBack if does not  designate public key. Therefore if client don't designate public key then server authenticate client by using PasswordCallback without password.

ci/citest/acceptance-test/tests/cmd_console_test.go
itouri: Added code to test the console. And change function name to TestCmdConsole_ShowOptionAuthenticationNone.

unakatsuo (approved):

ci/citest/acceptance-test/tests/cmd_console_test.go
```diff
@@ -21,6 +22,23 @@ func runConsoleCmd(instance_id string, t *testing.T) {
 	RunCmdAndExpectFail(t, "sh", "-c", fmt.Sprintf("openvdc console %s -- false", instance_id))
 }
 
+func runConsoleCmdWithPrivatekey(instance_id string, private_key_path string, t *testing.T) {
+	RunCmdAndReportFail(t, "openvdc", "console", instance_id, "-i", private_key_path)
+	RunCmdAndReportFail(t, "sh", "-c", fmt.Sprintf("openvdc console %s -i %s ls", instance_id, private_key_path))
```
Metallion: Why use sh here? Why not just do this?

```
RunCmdAndReportFail(t,"openvdc", "console", instance_id, "-i",  private_key_path, "ls")
```

Also you should run gofmt on this file and any other files that aren't formatted correctly. It's best to set up your editor to run it on save.

ci/citest/acceptance-test/tests/cmd_console_test.go
```diff
@@ -21,6 +22,23 @@ func runConsoleCmd(instance_id string, t *testing.T) {
...
+	RunCmdAndReportFail(t, "sh", "-c", fmt.Sprintf("openvdc console %s -i %s -- ls", instance_id, private_key_path))
+	RunCmdAndExpectFail(t, "sh", "-c", fmt.Sprintf("openvdc console %s -i %s -- false", instance_id, private_key_path))
+}
```
Metallion: You should also confirm that the console command fails when you provide the wrong private key.

Metallion (changes requested):
I think this is almost ok but it just needs a test to confirm that openvdc console fails when the wrong private key is provided.

ci/citest/acceptance-test/tests/cmd_console_test.go
```diff
@@ -22,11 +22,12 @@ func runConsoleCmd(instance_id string, t *testing.T) {
...
-	RunCmdAndReportFail(t, "openvdc", "console", instance_id, "-i", private_key_path)
-	RunCmdAndReportFail(t, "sh", "-c", fmt.Sprintf("openvdc console %s -i %s ls", instance_id, private_key_path))
-	RunCmdAndReportFail(t, "sh", "-c", fmt.Sprintf("openvdc console %s -i %s -- ls", instance_id, private_key_path))
-	RunCmdAndExpectFail(t, "sh", "-c", fmt.Sprintf("openvdc console %s -i %s -- false", instance_id, private_key_path))
+func runConsoleCmdWithPrivatekey(instance_id string, private_key_path string, t *testing.T, expect_fail bool) {
+	if expect_fail {
+		RunCmdAndExpectFail(t, "openvdc", "console", instance_id, "-i", private_key_path)
+	} else {
+		RunCmdAndReportFail(t, "openvdc", "console", instance_id, "-i", private_key_path)
+	}
```
Metallion: This function is only called twice. Once with expect_fail = true and once with expect_fail = false.

Since only one line gets executed for each call, I think it's better to write those lines directly in the test instead of in a separate function.

Metallion (changes requested):
Since the ESXi feature was merged to master, the tests started failing on this one.

```
# github.com/axsh/openvdc/scheduler
scheduler/mesos.go:174: cannot use esxi (type *model.EsxiTemplate) as type model.InstanceResource in argument to model.IsMatchingNodeGroups:
	*model.EsxiTemplate does not implement model.InstanceResource (missing GetAuthenticationType method)
```

I know this is annoying since it worked before, but you'll have to update to code to work with the latest master.

cmd/openvdc/cmd/console.go
```diff
@@ -71,8 +77,37 @@ var consoleCmd = &cobra.Command{
...
+				HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+			}
+
+			switch res.AuthType {
+			case model.AuthenticationType_NONE:
+				config.Auth = []ssh.AuthMethod{ssh.Password("")}
+			case model.AuthenticationType_PUB_KEY:
+				if indentityFile == "" {
+					log.Fatalf("Required private key but not setted")
+				} else {
```
unakatsuo: no need for the ``else`` block.

cmd/openvdc/cmd/console.go
```diff
@@ -19,8 +21,11 @@ import (
 	"google.golang.org/grpc"
 )
 
+var indentityFile string
```
unakatsuo: Typo and no need to be defined as a global variable. Can be refered as ``cmd.Flags().GetString("identity-file")``.

proto/v1.proto
```diff
@@ -90,6 +90,7 @@ message ConsoleReply {
   string instance_id = 1 [json_name="instance_id"];
   model.Console.Transport type = 2;
   string address = 3;
+  model.AuthenticationType authType = 4;
```
unakatsuo: Breaks the convention.

```
model.AuthenticationType auth_type = 4 [json_name="auth_type"];
```

unakatsuo (changes requested):

cmd/openvdc/cmd/console.go
itouri: Modified typo and using to GetString()

cmd/openvdc/cmd/console.go
itouri: Deleted it.

proto/v1.proto
itouri: Added json_name tag for auth_type.

ci/citest/acceptance-test/tests/cmd_console_test.go
itouri: Added TestLXCCmdConsole_AuthenticationPubkeymethod testing worng key.

ci/citest/acceptance-test/tests/cmd_console_test.go
itouri: Deleted runConsoleCmdWithPrivatekey and integrated to TestLXCCmdConsole_AuthenticationPubkey.

ci/citest/acceptance-test/tests/cmd_console_test.go
itouri: This test function integrate to TestLXCCmdConsole_AuthenticationPubkeymethod and modified to doesn't using "sh".

handlers/vm/lxc/lxc.go
```diff
@@ -85,13 +96,21 @@ func (h *LxcHandler) MergeArgs(dst model.ResourceTemplate, args []string) error
...
 	flags.IntVar(&mem, "memory_gb", int(mdst.MinMemoryGb), "")
+	defAuth := model.AuthenticationType_name[int32(mdst.AuthenticationType)]
+	flags.StringVar(&authType, "authentication_type", defAuth, "")
+	flags.StringVar(&sshPubkey, "ssh_public_key", mdst.SshPublicKey, "")
 	if err := flags.Parse(args); err != nil {
 		return err
 	}
 	mdst.Vcpu = int32(vcpu)
 	mdst.MemoryGb = int32(mem)
+	format := model.AuthenticationType_value[strings.ToUpper(authType)]
```
unakatsuo: Missing the validation  if ``--authtication_type`` gets unknown value.

unakatsuo (changes requested):

handlers/vm/lxc/lxc.go
itouri: Added validation of authtication_type to lxc.go and qemu.go.

PR #182, opened 2017-07-04
Feature qemu driver

toros11:
This PR implements a driver for running qemu instances on top of openvdc.
The CI environment has also been updated with two new nodes for qemu-executor to run on, one for linux bridges and one for open vSwitch.

Images that running through the qemu driver needs to have openvdc-init configured to boot as a service manually to be usable.


ci/citest/acceptance-test/multibox/10.0.100.16-vdc-executor-qemu/build.sh
```diff
@@ -0,0 +1,34 @@
...
+    "hosts"
+    "disable-firewalld"
+    "epel"
+    "qemu"
+    "kvm"
+    "ncat"
+    "bridge-utils"
+    "mesosphere-repo"
+    "mesos-agent"
+    "vdc-images"
```
Metallion, line 22: I think it is better to have only one image server in the CI environment instead of having it locally on multiple executors. That seems more likely in a production environment and tests our capability to get images from a remote location.

Edit: This is me assuming that we have a web server serving images for the executors. Please correct me if I'm wrong.

ci/citest/acceptance-test/multibox/ind-steps/step-kvm/common.source
```diff
@@ -0,0 +1,18 @@
...
+function nested_enabled {
+    local disabled=false
+
+    case ${ACCEL_TYPE} in
+        vmx) [[ "$(cat /sys/module/kvm/intel/parameters/nested)" == "Y" ]] || disabled=true ;;
+        svm) [[ $(cat /sys/module/kvm_amd/parameters/nested) -eq 1 ]] || disabled=true ;;
+    esac
+
+    return $disabled
+}
```
Metallion: The functions in this file don't seem to be used anywhere so why include them?

Also the top function has some typos. `kvm_enbabled` and `kmv`.

ci/citest/acceptance-test/multibox/10.0.100.17-vdc-executor-qemu-ovs/build.sh
```diff
@@ -0,0 +1,34 @@
...
+
+IND_STEPS=(
+    "box"
+    "ssh"
+    "hosts"
+    "disable-firewalld"
+    "epel"
+    "openvswitch"
+    "qemu"
+    "kvm"
```
Metallion: Why split qemu and kvm into two steps? I doubt we will ever install one without the other.

ci/citest/acceptance-test/multibox/ind-steps/step-vdc-images/init.sh
```diff
@@ -1,12 +1,20 @@
...
+    qemu) # stores directly in cache until we can use compressed files
+        IMAGES=( "${QEMU_IMAGES[@]}" )
+        LOCAL_STORAGE_PATH="${NODE_DIR}/guestroot/var/cache/qemu"
+        ;;
+    lxc) # stores on local image server
+        IMAGES=( "${LXC_IMAGES[@]}" )
+        LOCAL_STORAGE_PATH="${NODE_DIR}/guestroot/var/www/html/images"
+        ;;
+    *)
+        echo "Unknown image type."
```
Metallion: This should include `${image_type}` to make it a lot easier to debug. Also maybe we should fail when there's an unknown image type?

I figure this code should path ideally never be executed so if we fail more explicitly here, it'll make it easier to debug. An error message is easily missed in the log and it is likely to cause problems further down the line.

ci/citest/acceptance-test/tests/cmd_console_test.go
```diff
@@ -23,3 +20,15 @@ func TestCmdConsole_ShowOption(t *testing.T) {
...
+	stdout, _ := RunCmdAndReportFail(t, "openvdc", "run", "centos/7/lxc")
+	instance_id := strings.TrimSpace(stdout.String())
+	runConsoleCmd(instance_id, t)
+}
+
+// func TestQEMUCmdConsole_ShowOption(t *testing.T) {
+// 	stdout, _ := RunCmdAndReportFail(t, "openvdc", "run", "centos/7/qemu")
+// 	instance_id := strings.TrimSpace(stdout.String())
+// 	runConsoleCmd(instance_id, t)
+// }
```
Metallion, line 34: I take it qemu console is being developed in another branch that uncomments this code?

cmd/qemu-ifdown/main.go
```diff
@@ -0,0 +1,48 @@
...
+	"os"
+	"os/exec"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/spf13/viper"
+)
+
+var DefaultConfPath string
+
+// todo: error handling
```
Metallion: That's one scary todo if left unresolved. :p

Metallion (changes requested):
Reviewed the CI/test code. Will need @unakatsuo's approval for the main code.

ci/citest/acceptance-test/multibox/10.0.100.16-vdc-executor-qemu/build.sh
toros11: The reason I decided to include them is because the driver does not support compressed types right now and it took awhile to download them image. The "local image server" also seems like it doesn't translate as well to qemu, the images doesn't really have a template and I'm not so sure about parsing out a "local" keyword just for that sake, instead just supplying the full url (local ip) in the json file feels more natural to me.

ci/citest/acceptance-test/multibox/10.0.100.17-vdc-executor-qemu-ovs/build.sh
toros11: It was because I mistook nested functionality with kvm module getting loaded, theoretically qemu executor does not require kvm but I can merge these if that makes more sense.

ci/citest/acceptance-test/tests/cmd_console_test.go
toros11: Correct

cmd/qemu-ifdown/main.go
toros11: I wasn't sure how much error handling was required here, it seems like the config itself has different standards that I didn't realize when I created this, so there may not be much error handling to do after all.

ci/citest/acceptance-test/multibox/ind-steps/step-kvm/common.source
toros11: At first I was not sure wether the scripts should try to enable kvm/the nested parameter or not, opinions on this? This also ties into the comment about kvm/qemu being separate steps.
I enabled nested manually and I'm not even sure it can be done from within the docker container

hypervisor/qemu/qemu.go
```diff
@@ -0,0 +1,368 @@
...
+}
+
+func renderData(keyPath string, key string, value interface{}) {
+	switch value.(type) {
+	case string:
+		ioutil.WriteFile(filepath.Join(keyPath, key), []byte(value.(string)), 0644)
+		return
+	default:
+		kp := filepath.Join(keyPath, key)
+		os.MkdirAll(kp, os.ModePerm)
```
unakatsuo: missing error check.

hypervisor/qemu/qemu.go
```diff
@@ -0,0 +1,368 @@
...
+			}
+		}
+		return metadataMap
+	})
+}
+
+func (d *QEMUHypervisorDriver) CreateInstance() error {
+	d.log().Infoln("Create instance...")
+	instanceDir := filepath.Join(settings.InstancePath, d.Base.Instance.GetId())
+	os.MkdirAll(instanceDir, os.ModePerm)
```
unakatsuo: missing error check.

hypervisor/qemu/qemu.go
```diff
@@ -0,0 +1,368 @@
...
+
+	instanceImage := d.machine.Drives["disk"].Image
```
unakatsuo: The map item should be checked for nil case.

```
drive, ok := d.machine.Drives["disk"]
if !ok {
  return error
}
```

hypervisor/qemu/qemu.go
```diff
@@ -0,0 +1,368 @@
...
+		}
+		if err := instanceImage.SetBaseImage(baseImage); err != nil {
+			return err
+		}
+		if err := instanceImage.CreateImage(); err != nil {
+			return err
+		}
+	}
+
+	metadriveImage := d.machine.Drives["meta"].Image
```
unakatsuo: The map item should be checked for nil case.

hypervisor/qemu/qemu_machine.go
```diff
@@ -0,0 +1,215 @@
...
+	return fmt.Sprintf("openvdc@%s", m.Name)
+}
+
+func (m *Machine) HavePrompt() bool {
+	c, err := net.Dial("unix", m.Serial)
+	if err != nil {
+		return false
+	}
+
+	matchprompt := make(chan bool, 1)
```
unakatsuo: missing to close the go channel.

```
defer close(matchprompt)
```

hypervisor/qemu/qemu_machine.go
```diff
@@ -0,0 +1,215 @@
...
+		time.Sleep(timeout)
+		timeoutc <- true
+	}()
+
+	go func() {
+		for {
+			select {
+			case <-timeoutc:
+				errorc <- errors.Errorf("Timed out scheduling state %s", stateValues[nextState])
+				return
```
unakatsuo, line 140: https://golang.org/pkg/time/#After

ci/citest/acceptance-test/multibox/10.0.100.17-vdc-executor-qemu-ovs/build.sh
ChrisSalisbury: Just my 2 cents, but qemu can run lots of things that we might implement in the future at some point (like xen). So it might make sense to leave them separated especially since it's already done that way...

handlers/vm/qemu/qemu.go
```diff
@@ -0,0 +1,126 @@
...
+	if tmpl.GetQemuImage() == nil {
+		return nil, handlers.ErrInvalidTemplate(h, "qemu_image must exist")
+	}
+
+	return tmpl, nil
+}
+
+func (h *QemuHandler) SetTemplateItem(t *model.Template, m model.ResourceTemplate) {
+	t.Item = &model.Template_Qemu{
+		Qemu: m.(*model.QemuTemplate),
```
unakatsuo: Type assertion must be done earlier. It can panic if unexpected resource template was passed.

```
qemu_tmpl, ok :=  m.(*model.QemuTemplate)
if !ok {
  panic(".....")
}

t.Item = &model.Template_Qemu{
		Qemu: qemu_tmpl,
}

```

hypervisor/qemu/qemu.go
```diff
@@ -0,0 +1,368 @@
...
+	settings.ImageServerUri = u
+	settings.CachePath = sub.GetString("hypervisor.cache-path")
+	settings.InstancePath = sub.GetString("hypervisor.instance-path")
+	return nil
+}
+
+func (p *QEMUHypervisorProvider) CreateDriver(instance *model.Instance, template model.ResourceTemplate) (hypervisor.HypervisorDriver, error) {
+	qemuTmpl, ok := template.(*model.QemuTemplate)
+	if !ok {
+		return nil, errors.Errorf("template type is not *model.QemuTemplate: %T, template")
```
unakatsuo: Wrong double quote position.

```
errors.Errorf("template type is not *model.QemuTemplate: %T", template)
```

hypervisor/qemu/qemu.go
```diff
@@ -0,0 +1,368 @@
...
+			BridgeHelper: settings.QemuBridgeHelper,
+		})
+	}
+
+	d.machine = NewMachine(int(d.template.GetVcpu()), uint64(d.template.GetMemoryGb()*1024))
+	d.machine.Drives["disk"] = Drive{Image: NewImage(filepath.Join(instanceDir, "diskImage."+imageFormat), imageFormat)}
+	d.machine.Drives["meta"] = Drive{Image: NewImage(filepath.Join(instanceDir, "metadrive.img"), "raw"), If: "floppy"}
+	d.machine.Name = instanceId
+	d.machine.Monitor = fmt.Sprintf("%s", filepath.Join(instanceDir, "monitor.socket"))
+	d.machine.Serial = fmt.Sprintf("%s", filepath.Join(instanceDir, "serial.socket"))
```
unakatsuo: No need for ``fmt.Sprintf()``

```
d.machine.Monitor =filepath.Join(instanceDir, "monitor.socket")
d.machine.Serial = filepath.Join(instanceDir, "serial.socket")
```

hypervisor/qemu/qemu.go
```diff
@@ -0,0 +1,368 @@
...
+		for key, value := range value.(map[string]interface{}) {
+			renderData(kp, key, value)
+		}
+	}
+}
+
+func (d *QEMUHypervisorDriver) addMetadata(metadataDrive *Image, datamap func(machine *Machine) map[string]interface{}) error {
+	mountPath := filepath.Join(filepath.Dir(metadataDrive.Path), "meta-data")
+
+	os.MkdirAll(mountPath, os.ModePerm)
```
unakatsuo: missing eeror check.

hypervisor/qemu/qemu_cmd.go
```diff
@@ -0,0 +1,82 @@
...
+	if len(m.Pidfile) > 0 {
+		cmd.appendArgs("-pidfile", m.Pidfile)
+	}
+	for _, d := range m.Drives {
+		drive := fmt.Sprintf("file=%s,format=%s", d.Image.Path, d.Image.Format)
+		if len(d.If) > 0 {
+			drive = fmt.Sprintf("%s,if=%s", drive, d.If)
+		}
+		cmd.appendArgs("-drive", drive)
+	}
```
unakatsuo, line 41: How do you manage the drive device order? It seems not to be deterministic if the OS drive has earlier drive number.

hypervisor/qemu/qemu_machine.go
```diff
@@ -0,0 +1,215 @@
...
+
+type Machine struct {
+	State   State
+	Cores   int
+	Memory  uint64
+	Name    string
+	Display string
+	Vnc     string
+	Monitor string
+	Serial  string
```
unakatsuo: If they represent UNIX socker path they should be renamed.

```
MonitorSocketPath string
SerialSocketPath string
```

unakatsuo (changes requested):

hypervisor/qemu/qemu_cmd.go
unakatsuo: > According to the Go spec, the order of iteration over a map is undefined, and may vary between runs of the program.

https://stackoverflow.com/a/23330883

cmd/qemu-ifdown/main.go
```diff
@@ -0,0 +1,48 @@
...
+	c.Run()
+}
+
+func init() {
+	viper.SetDefault("bridges.type", "linux")
+	viper.SetDefault("bridges.name", "br0")
+	viper.SetConfigFile(DefaultConfPath)
+	viper.SetConfigType("toml")
+	viper.AutomaticEnv()
+	err := viper.ReadInConfig()
```
unakatsuo, line 29: Basically ``func init()`` is not the place to make OS syscall operation. Please move to ``main()``.

cmd/qemu-ifup/main.go
```diff
@@ -0,0 +1,48 @@
...
+	c.Run()
+}
+
+func init() {
+	viper.SetDefault("bridges.type", "linux")
+	viper.SetDefault("bridges.name", "br0")
+	viper.SetConfigFile(DefaultConfPath)
+	viper.SetConfigType("toml")
+	viper.AutomaticEnv()
+	err := viper.ReadInConfig()
```
unakatsuo, line 29: Basically func init() is not the place to make OS syscall operation. Please move to main().

unakatsuo (changes requested):

cmd/qemu-ifdown/main.go
unakatsuo: At least all the errors from ``cmd.Run()`` must be caught. Running external program has many error cases so that it will make debugging integration test harder.

cmd/qemu-ifup/main.go
toros11: Fixed

cmd/qemu-ifdown/main.go
toros11: Fixed

hypervisor/qemu/qemu_cmd.go
toros11: There is actually no management of the device driver order right now, we only have the os drive and a floppy attached, and the floppy is not set to boot.

hypervisor/qemu/qemu_machine.go
toros11: Fixed

hypervisor/qemu/qemu_machine.go
```diff
@@ -0,0 +1,216 @@
...
+		return false
+	}
+	b, _ := buf.ReadBytes('\n')
+	return (strings.Contains(string(b), m.promptPattern()))
+}
+
+// since machine struct does not get saved in memory for each instance there may not be any points
+// in scheduling states as they are not stored anywhere
+func (m *Machine) ScheduleState(nextState State, timeout time.Duration, evaluation func() bool) error {
+	errorc := make(chan error)
```
unakatsuo, line 130: missing to close go channel.

unakatsuo (changes requested):

hypervisor/qemu/qemu_machine.go
```diff
@@ -0,0 +1,216 @@
...
+	}()
+
+	// send new line in order to trigger the prompt
+	fmt.Fprintf(c, "\n")
+	return <-matchprompt
+}
+
+func (m *Machine) WaitForPrompt() bool {
+	c, err := net.Dial("unix", m.SerialSocketPath)
+	defer c.Close()
```
unakatsuo: This defer should come below the ``if err != nil``.

unakatsuo (changes requested):

hypervisor/qemu/qemu_cmd.go
unakatsuo: Can be improved in later change.

hypervisor/qemu/qemu_machine.go
toros11: Fixed, however channels are garbage collected and only needs to be closed when the receiver needs to know that no more data is to be sent on it.

https://groups.google.com/d/msg/golang-nuts/pZwdYRGxCIk/rGr1D-uTQMEJ

hypervisor/qemu/qemu_machine.go
toros11: > Fixed

hypervisor/qemu/qemu_machine.go
unakatsuo: Good to know that go channel is GCed. From my experience, the closing explicitly helps to find logical issue of channels because the writing or reading to the closed channel causes panic.
  
It is easy to write the code that causes leaking goroutine and channel.

https://stackoverflow.com/questions/38090545/are-goroutines-garbage-collected-together-with-their-channels

unakatsuo (approved):

ci/citest/acceptance-test/multibox/10.0.100.17-vdc-executor-qemu-ovs/build.sh
Metallion: In my opinion that would be a premature optimization. I don't think we're likely to implement Xen or other qemu features any time soon so I would merge the steps to keep the code simple.

We don't always have to do things my way though so if @toros11 agrees with you, I'm ok with leaving it this way.

Metallion (changes requested):
Forgot one thing. Since new nodes are added to the acceptance test, the illustration here needs to be updated to match.

https://github.com/axsh/openvdc/tree/master/ci/citest/acceptance-test/illustrations

ci/citest/acceptance-test/multibox/ind-steps/step-qemu-kvm/install.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/bash
 
+# install_yum_package "qemu-system-x86"
```
Metallion: Why is this commented out? Perhaps because it gets installed as a dependency? Personally I'm not a fan of including commented code since old code is preserved in the git history anyway but not gonna hold back on merging for that.

ci/citest/acceptance-test/multibox/ind-steps/step-qemu-kvm/postconfigure.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+procc_type=
+case ${ACCEL_TYPE} in
+    vmx) procc_type="intel" ;;
+    svm) procc_type="amd" ;;
+    *) echo "Invalid hardware acceleration" ; exit 255 ;;
```
Metallion, line 7: I would suggest using a different exit code than 255 because 255 has the special meaning: "exit code out of range." http://tldp.org/LDP/abs/html/exitcodes.html

It's not a big deal and can be fixed in a later change though.

Unrelated side note: I remembered that 255 was special but when looking up what it means exactly, I also learned that exit codes higher than 255 will return the result of (\<exit code> mod 256). That means any exit code divisible by 256 will return exit code 0. :open_mouth:

toros11: Bashsteps sends exit code 255 when it exits so the loop that iterates the steps listens for it, since this is not inside a step I forced the exit code, maybe these type of checks should be put inside a step as well though.

Metallion: I see. :) Didn't know that.

Metallion (approved):

PR #184, opened 2017-07-05
Update docs after local img merge

Metallion:
* The local images feature added an `IMG_DIR` variable to the config file that wasn't added to the readme. Fixed that.

* Fixed a variable's default value that was accidentally changed

* Changed the illustration in the readme from an embed to a link since SVG files aren't rendered in embeds for security reasons.


unakatsuo (approved):

unakatsuo:
Sometime later in this year we may be able to embed .svg to README.md.

https://github.com/github/markup/issues/556

Metallion:
I see. So maybe reverting that commit would be better so it just starts working once github's fix (workaround?) comes through.

unakatsuo:
you can proceed to merge.

PR #191, opened 2017-07-21
Feature qemu console

toros11:
This adds console feature to the qemu driver.
Currently the interactive console is represented as a serial console, while executing commands
is done by using the qemu guest agent which needs to be installed on the image.
Piped commands are currently not supported.


hypervisor/qemu/qemu.go
```diff
@@ -385,6 +406,6 @@ func (d QEMUHypervisorDriver) RebootInstance() error {
 	})
 }
 
-func (d QEMUHypervisorDriver) InstanceConsole() hypervisor.Console {
-	return nil
-}
+// func (d QEMUHypervisorDriver) InstanceConsole() hypervisor.Console {
```
unakatsuo: How do you implement the console feature without this method?

unakatsuo (changes requested):

hypervisor/qemu/qemu.go
toros11: Oh this comment should be removed, it's doesn't get the pointer receiver.

The correct method is implemented here:
https://github.com/axsh/openvdc/blob/19a4b597f27b9db7f983dbcda81426f9ac25242c/hypervisor/qemu/qemu_console.go#L196

unakatsuo (approved):

PR #194, opened 2017-07-27
Fix local image tar issue

Metallion:
Mholt's archiver library that we were using to unpack the LXC rootfs tarball wasn't quite getting the job done. A rootfs contains a bunch of device nodes and other special files in the `/dev` directory. This library couldn't handle those properly and just wrote empty files instead.

I've checked out a couple other implementations but most seemed to just read the tarball, iterate over the files in it and then write new files with the same name and contents. While this works for the most basic archives, tarballs keep a lot more info such as ownership, permissions, etc.

I discussed with @unakatsuo and we decided to just use the tar command directly as there seems to be no sufficient go implementation and it's a safe bet that this lxc template will never be used outside of Linux.

### DIFF
--- a/cmd/lxc-openvdc/main.go
+++ b/cmd/lxc-openvdc/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"flag"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


unakatsuo (changes requested):
Missing changes to the rpmspec file.

```
Requires: tar, xz, gzip
```

Metallion:
Oops, sorry about that. Will fix.

Edit: Why require gzip though? Currently only tar.xz archives are used.

unakatsuo (approved):

b0r6 (approved):

PR #196, opened 2017-08-01
Esxi driver

b0r6:
The executor needs a ssh-key to be present in `/etc/openvdc/ssh/esxi` to be able to access vmkfstools on the esxi-machine, which it uses to clone vmdks. For example, running openvdc run centos/7/esxi clones a base CentOS7 vm into one called i-000000001

The executor waits for the instance vm to get an ip.  When it has gotten one, the executor sets the state to RUNNING. This was a simple way for the executor to make sure that the instance was actually up and running, but the downside is that it requires an ip to be set, so at some later point it will probably use a different method for determining this.
`openvdc run centos/7/esxi '{"interfaces":[{"Ipv4Addr":"10.0.100.80"}]}'`


hypervisor/esxi/esxi.go
```diff
@@ -0,0 +1,309 @@
...
+	a = append(a, args[0])
+	a = append(a, fmt.Sprintf("-dc=%s", settings.EsxiDatacenter))
+	a = append(a, fmt.Sprintf("-k=%s", strconv.FormatBool(settings.EsxiInsecure)))
+	a = append(a, fmt.Sprintf("-u=%s", settings.EsxiUrl))
+
+	for i := 1; i < len(args); i++ {
+		a = append(a, args[i])
+	}
+
+	cli.Run(a)
```
unakatsuo: How can we handle errors from ``cli.Run``? All the lines where ``esxiCmd()`` is called are missing to handle error.

hypervisor/esxi/esxi.go
```diff
@@ -0,0 +1,309 @@
...
+	return d.Base.Log
+}
+
+func esxiCmd(args ...string) {
+	var a []string
+
+	a = append(a, args[0])
+	a = append(a, fmt.Sprintf("-dc=%s", settings.EsxiDatacenter))
+	a = append(a, fmt.Sprintf("-k=%s", strconv.FormatBool(settings.EsxiInsecure)))
+	a = append(a, fmt.Sprintf("-u=%s", settings.EsxiUrl))
```
unakatsuo: Here can be rewrite:

```
a := []string {
args[0],
fmt.Sprintf("-dc=%s", settings.EsxiDatacenter),
fmt.Sprintf("-k=%s", strconv.FormatBool(settings.EsxiInsecure)),
fmt.Sprintf("-u=%s", settings.EsxiUrl),
}
```

hypervisor/esxi/esxi_console.go
```diff
@@ -0,0 +1,145 @@
...
+				if bytes.Contains(b[0:n], []byte{0x11}) {
+					c.conChan <- nil
+				}
+				_, err = param.Stdout.Write(b[0:n])
+				if err != nil {
+					c.conChan <- errors.Wrap(err, "\nFailed to write to the console stdout buffer\n\n")
+				}
+			}
+		}
+		defer waitClosed.Done()
```
unakatsuo: ``defer`` should come to the beginning of the ``go func()``.

hypervisor/esxi/esxi_console.go
```diff
@@ -0,0 +1,145 @@
...
+					log.Info("Received exit from stdin")
+					c.conChan <- errors.Wrap(err, "\nConsole exited by ctrl-q\n\n")
+				}
+				_, err = c.telnetConn.Write(b[0:n])
+				if err != nil {
+					c.conChan <- errors.Wrap(err, "\nFailed to write to telnet connection from the buffer\n\n")
+				}
+			}
+		}
+		defer waitClosed.Done()
```
unakatsuo: ``defer`` should come to the beginning of the ``go func()``.

hypervisor/esxi/esxi.go
```diff
@@ -0,0 +1,309 @@
...
+	settings.ScriptPath = sub.GetString("hypervisor.script-path")
+	settings.EsxiUser = sub.GetString("hypervisor.esxi-user")
+	settings.EsxiPass = sub.GetString("hypervisor.esxi-pass")
+	settings.EsxiIp = sub.GetString("hypervisor.esxi-ip")
+	settings.EsxiDatacenter = sub.GetString("hypervisor.esxi-datacenter")
+	settings.EsxiInsecure = sub.GetBool("hypervisor.esxi-insecure")
+	settings.EsxiHostSshkey = sub.GetString("hypervisor.esxi-host-sshkey")
+	settings.EsxiVmUser = sub.GetString("hypervisor.esxi-vm-user")
+	settings.EsxiVmPass = sub.GetString("hypervisor.esxi-vm-pass")
+	settings.EsxiVmDatastore = sub.GetString("hypervisor.esxi-vm-datastore")
```
unakatsuo: Should validate the required parameters and then return error if they were blank.

hypervisor/esxi/esxi.go
```diff
@@ -0,0 +1,309 @@
...
+
+	esxiInfo := fmt.Sprintf("%s:%s@%s", settings.EsxiUser, settings.EsxiPass, settings.EsxiIp)
+	u, _ := url.Parse("https://")
+	u.Path = path.Join(u.Path, esxiInfo, "sdk")
```
unakatsuo: Let ``url.Parse()`` validate the syntax and return the format error.

```
u, err := url.Parse("https://" + esxiInfo + "/sdk")
if err != nil {
  return errors.Wrap(err, "Failed to parse url for ESXi server")
}
```

unakatsuo (changes requested):
Please add error check to all ``esxiCmd()`` calls and remove ``// TODO: error checks`` comments.

hypervisor/esxi/esxi.go
toros11: The `cli.Run()` function returns an int, `0` on success, `1` on failure. The error message is printed directly to os.Stderr by the function itself. 

The options here seems to be to either return an error from the `esxiCmd()` if the return code was `1`, 
or return the function call itself and let the calling function return the error based on the return code

unakatsuo (approved):

ci/citest/acceptance-test-esxi/tests/01_start_and_terminate_esxi_instance_test.go
```diff
@@ -0,0 +1,24 @@
...
+	stdout, _ := tests.RunCmdAndReportFail(t, "openvdc", "run", "centos/7/esxi", `{"interfaces":[{"Type":"veth","Ipv4Addr":"192.168.2.123"}]}`)
+	instance_id := strings.TrimSpace(stdout.String())
+
+	_, _ = tests.RunCmdAndReportFail(t, "openvdc", "show", instance_id)
+	tests.WaitInstance(t, 5*time.Minute, instance_id, "RUNNING", []string{"QUEUED", "STARTING"})
+
+	_, _ = tests.RunCmdWithTimeoutAndReportFail(t, 10, 5, "openvdc", "destroy", instance_id)
+	tests.WaitInstance(t, 5*time.Minute, instance_id, "TERMINATED", nil)
+
+	//TODO: Use Esxi api to make sure that the vm gets properly created and destroyed.
```
Metallion, line 23: Why not do this before merge? I'm afraid this will be forgotten about and might come back to bite us otherwise.

Metallion (commented):
Just out of curiosity I was wondering why we're waiting until the instance has an IP before we label it as running. For LXC and KVM we set it to running as soon as the KVM process/LXC container starts.

Metallion:
I know @unakatsuo reviewed the code before but since there have been new commits since then, maybe he should have another look?

ci/citest/acceptance-test-esxi/tests/01_start_and_terminate_esxi_instance_test.go
b0r6: As for the ip wait command: This used to be necessary, but indeed not anymore now after a few changes. I'll remove it.

unakatsuo (changes requested):
Please import ``github.com/pkg/errors`` and wrap the ``err`` variable returned from std libs. Some lines where receive ``err`` squash the original ``err``.

```
if err := os.MkDir("/....."); err != nil {
  return errors.Wrapf(err, "message %s", v1)
}
```

hypervisor/esxi/esxi.go
```diff
@@ -227,15 +250,121 @@ func (d *EsxiHypervisorDriver) vmPath() string {
...
 
+func renderData(keyPath string, key string, value interface{}) error {
+	switch value.(type) {
+	case string:
+		ioutil.WriteFile(filepath.Join(keyPath, key), []byte(value.(string)), 0644)
+		return nil
+	default:
+		kp := filepath.Join(keyPath, key)
+		if err := os.MkdirAll(kp, os.ModePerm); err != nil {
+			return errors.Errorf("Unable to create folder: %s", kp)
```
unakatsuo: wrap ``err``

hypervisor/esxi/esxi.go
```diff
@@ -227,15 +250,121 @@ func (d *EsxiHypervisorDriver) vmPath() string {
 	return join(0, "-vm.path=[", settings.EsxiVmDatastore, "]", storageImg(d.vmName))
 }
 
+func renderData(keyPath string, key string, value interface{}) error {
+	switch value.(type) {
+	case string:
+		ioutil.WriteFile(filepath.Join(keyPath, key), []byte(value.(string)), 0644)
```
unakatsuo:
```
return errors.Wrap(ioutil.WriteFile(filepath.Join(keyPath, key), []byte(value.(string)), 0644))
```

hypervisor/esxi/esxi.go
```diff
@@ -194,6 +209,14 @@ func join(separator byte, args ...string) string {
 	return buf.String()
 }
 
+func runCmd(cmd string, args []string) error {
+	c := exec.Command(cmd, args...)
+	if err := c.Run(); err != nil {
+		return errors.Errorf("failed to execute command :%s %s", cmd, args)
```
unakatsuo: Wrap ``err``.

handlers/vm/esxi/esxi.go
```diff
@@ -0,0 +1,106 @@
...
+
+	type EsxiImage struct {
+		DownloadUrl string `json:"download_url,omitempty"`
+	}
+
+	var json_template struct {
+		EsxiImage EsxiImage `json:"esxi_image,omitempty"`
+	}
+	if err := json.Unmarshal(in, &json_template); err != nil {
+		return nil, err
```
unakatsuo: Wrap ``err``

handlers/vm/esxi/esxi.go
```diff
@@ -0,0 +1,106 @@
...
+	if !ok {
+		return handlers.ErrMergeDstType(new(model.EsxiTemplate), dst)
+	}
+
+	flags := flag.NewFlagSet("esxi template", flag.ContinueOnError)
+	var vcpu, mem int
+	flags.IntVar(&vcpu, "vcpu", int(mdst.MinVcpu), "")
+	flags.IntVar(&mem, "memory_gb", int(mdst.MinMemoryGb), "")
+	if err := flags.Parse(args); err != nil {
+		return err
```
unakatsuo: Wrap ``err``.

hypervisor/esxi/esxi_console.go
```diff
@@ -0,0 +1,185 @@
...
+	// the exec command from sshd includes /bin/sh -c here, which is not compatible with the guest.run command
+	for _, arg := range strings.Split(args[2], " ") {
+		cmd = append(cmd, arg)
+	}
+
+	waitClosed.Add(1)
+	go func () {
+		rOut, wOut, err := os.Pipe()
+		if err != nil {
+			log.Info("failed os.Pipe for stdout")
```
unakatsuo: ``log.Warn()`` with showing ``err`` contents.

hypervisor/esxi/esxi_console.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+		stdout := os.Stdout // save original stdout
+		os.Stdout = wOut
+
+		rErr, wErr, err := os.Pipe()
+		if err != nil {
+			log.Info("failed os.Pipe for stderr")
```
unakatsuo: ``log.Warn()`` with showing ``err`` contents.

unakatsuo (changes requested):

hypervisor/util/serial_connection.go
```diff
@@ -0,0 +1,78 @@
...
+	go func() {
+		defer waitClosed.Done()
+		b := make([]byte, 8192) // 8 kB is the default page size for most modern file systems
+		for {
+			select {
+			case err := <-errc:
+				 errc <- err
+				break
+			default:
+				n, err := param.Stdin.Read(b)
```
unakatsuo, line 31: Missing handle ``io.EOF`` error case returned from ``.Read()``.

https://golang.org/pkg/io/#Reader

hypervisor/util/serial_connection.go
```diff
@@ -0,0 +1,78 @@
...
+					errc <- errors.Wrap(err, "\nFailed to read the connection buffer from socket ")
+				}
+				// exit on ctrl + q
+				if bytes.Contains(b[0:n], []byte{0x11}) {
+					errc <- nil
+				}
+				_, err = param.Stdout.Write(b[0:n])
+				if err != nil {
+					errc <- errors.Wrap(err, "\nFailed to write to the console stdout buffer\n\n")
+				}
```
unakatsuo, line 85: Might be good to flush the Stdout buffer?

```
if synchable, ok := param.Stdout.(*os.File); ok {
	if err := synchable.Sync(); err != nil {
		log.Warn("Failed %v to flush: %s", param.Stdout, err)
	}
}
```

unakatsuo (changes requested):

hypervisor/util/serial_connection.go
```diff
@@ -28,7 +28,11 @@ func (sc *SerialConnection) AttachSerialConsole(param *hypervisor.ConsoleParam,
 			default:
 				n, err := param.Stdin.Read(b)
 				if err != nil {
-					errc <- errors.Wrap(err, "\nFailed to read from the from the console input buffer\n\n")
+					if err == io.EOF {
+						errc <- errors.Wrap(nil, "\nConsole exited by EOF\n\n")
```
unakatsuo: The error message at second argument will never be sent back as error.

https://github.com/pkg/errors/blob/master/errors.go#L181-L183

Print as a log message if you need to catch.

unakatsuo (changes requested):

unakatsuo (approved):

PR #205, opened 2017-10-23
Properly update connection status

toros11:
Due to how protobuf handles enums the connection status did not get updated properly as the 0th element always defaults to null. This PR adds an empty value as the 0th element to mark that it's not intended to be used.

I also refactored the acceptance test to not use sleeps and instead wait for the expected connection statuses, this should solve the issues with occasionally failing tests.


b0r6 (approved):

unakatsuo (approved):

Metallion (approved):
Awesome. That acceptance test refactor was on my own todo list as well. :)

PR #208, opened 2017-11-14
Copy command

b0r6:
`./openvdc copy [filepath] [instanceID]:[path]`

As of now this can only copy a local file and transfer it to the instance machine, not the other way around. Ideally one should also be able to copy files between instances as well, but I'll implement that at some later point.


unakatsuo (changes requested):
missing ``go fmt``.

cmd/openvdc/cmd/copy.go
```diff
@@ -0,0 +1,76 @@
+package cmd
+
+import (
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/axsh/openvdc/api"
+	filecopy "github.com/axsh/openvdc/cmd/openvdc/cmd/copy"
```
unakatsuo, line 8: weird namespace.

cmd/openvdc/cmd/copy/copy.go
```diff
@@ -0,0 +1,84 @@
...
+		return true
+	}
+}
+
+func (c *Client) CopyFile(filePath string, instanceDir string) error {
+	if !fileExists(filePath) {
+		return errors.New("Source file not found")
+	}
+
+	file, _ := os.Open(filePath)
```
unakatsuo: Please do not squash the error from IO operation.

cmd/openvdc/cmd/copy/copy.go
```diff
@@ -0,0 +1,84 @@
...
+	session, err := cl.NewSession()
+	if err != nil {
+		return errors.Wrap(err, "Failed to create session")
+	}
+	defer session.Close()
+
+	go func() {
+		w, _ := session.StdinPipe()
+		defer w.Close()
+		fmt.Fprintln(w, "C0655", int64(len(b)), srcFilename)
```
unakatsuo: Here can rewrite with https://golang.org/pkg/os/#File.Stat. No need the buffering the content at L57-58. It is scary for the large file.

cmd/openvdc/cmd/copy/copy.go
```diff
@@ -0,0 +1,84 @@
...
+		io.Copy(w, br)
```
unakatsuo: Just pass ``file`` at L54. And please do not forget to handle the error from the function.

```
io.Copy(w, file)
```

cmd/openvdc/cmd/copy/copy.go
```diff
@@ -0,0 +1,84 @@
...
+	if !fileExists(filePath) {
+		return errors.New("Source file not found")
+	}
+
+	file, _ := os.Open(filePath)
+	defer file.Close()
+
+	b, _ := ioutil.ReadAll(file)
+	br := bytes.NewReader(b)
+	srcFilename := path.Base(filePath)
```
unakatsuo: ``path`` -> ``filepath``

cmd/openvdc/cmd/copy/copy.go
```diff
@@ -0,0 +1,84 @@
...
+		return false
+	} else {
+		return true
+	}
+}
+
+func (c *Client) CopyFile(filePath string, instanceDir string) error {
+	if !fileExists(filePath) {
+		return errors.New("Source file not found")
+	}
```
unakatsuo: ``filesExists()`` is redundant.

```
file, err := os.Open(filePath)
if err != nil {
  if os.IsNotExist(err) {
    return errors.New("Source file not found")
  }
  return errors.Wrap(err, "Open failed")
}
defer file.Close()
```

unakatsuo (changes requested):

unakatsuo (changes requested):
I guess this change only works for LXC hypervisor. It should return error if the destination is the unsupported hypervisor type.

cmd/openvdc/cmd/copy.go
b0r6: Ok. Any suggestion?

unakatsuo: I did same hack for console. please ignore.

cmd/openvdc/cmd/copy/copy.go
```diff
@@ -0,0 +1,87 @@
...
+
+	session, err := cl.NewSession()
+	if err != nil {
+		return errors.Wrap(err, "Failed to create session")
+	}
+	defer session.Close()
+
+	go func() {
+		w, _ := session.StdinPipe()
```
unakatsuo: Squashing error.

cmd/openvdc/cmd/copy/copy.go
```diff
@@ -0,0 +1,87 @@
...
+		defer w.Close()
+		fmt.Fprintln(w, "C0655", int64(fileinfo.Size()), srcFilename)
+		io.Copy(w, file) //Todo: handle error
```
unakatsuo: Please handle the error.

unakatsuo (changes requested):

cmd/openvdc/cmd/copy/copy.go
```diff
@@ -0,0 +1,87 @@
...
+		fmt.Fprintln(w, "\x00")
+	}()
+
+	session.Run("/usr/bin/scp -t " + instanceDir)
```
unakatsuo: Squashing error.

unakatsuo (changes requested):

Metallion:
Looks good to me. If @unakatsuo  is happy with the changes, then I think this is good to go. :+1:

Metallion (approved):

cmd/openvdc/cmd/copy.go
```diff
@@ -0,0 +1,76 @@
...
+		if err != nil {
+			log.WithError(err).Fatal("Disconnected abnormally")
+		}
+
+		client, err := filecopy.NewClient(res)
+		if err != nil {
+			log.WithError(err).Fatal("Failed to create client")
+		}
+
+		err = client.CopyFile(src, instanceDir)
```
unakatsuo, line 68: Does ``instanceDir`` point to the path on the host filesystem? It does not make sense. This feature is to copy the file into the instance filesystem not to the host filesystem.

unakatsuo (changes requested):

cmd/openvdc/cmd/copy.go
b0r6: Example:
```./openvdc copy testfile i-0000000001:/tmp/```

instanceDir refers to whichever folder you want to put the file in on the instance filesystem, in this case "/tmp/".

The file ends up in ```"/var/lib/lxc/i-0000000001/rootfs/tmp/"``` not ```"/tmp/"```

File: cmd/openvdc/cmd/copy/copy.go [hunk 11]
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	return &Client{
+		ClientConfig: config,
+		Host:         host,
+		Port:         port,
+	}, nil
+}
+
+func fileExists(path string) bool {
```
unakatsuo: You can remove this unused function.

unakatsuo (changes requested):

cmd/openvdc/cmd/copy.go
unakatsuo: Miss read the code. ``/usr/bin/scp`` is run by executor's sshd so that data stream is sent to the container.

unakatsuo (approved):

PR #210, opened 2017-11-15
Fix serial connection

toros11:
Fixed some  problems with the serial console

* Executor failed due to nil pointer.
The goroutine reading the connection would sometimes be left running after the connection had already been closed resulting in the executor crashing.

* The return value sent returned by wait did not implement correct interface type.
The ssh service running on the executor expects the errors to be type asserted into `ConsoleWaitError` type.

* Exiting the console with return code, did not return a new line


unakatsuo (approved):
It looks ok as a fix PR. But I imagine that they can  simplify by rewriting with https://golang.org/pkg/io/#Copy or https://golang.org/pkg/io/#Pipe and Reader/Writer wrapper for some control codes.

PR #211, opened 2017-11-17
Storable scheduler

itouri:
New openvdc scheduler storing offer of mesos slave noticed by mesos master.
When coming request from user, scheduler refer to stored offers and judging whether mesos slave satisfing request is exist.
When exist mesos slave satisfing request, accept the request. If it does not exist, scheduler informs the user that there are no slaves that satisfy the request.


unakatsuo (changes requested):
``model/store.go`` should not be a part of model class. This should be a private struct of ``scheduler`` or ``mesos`` package.

handlers/vm/lxc/lxc_test.go
```diff
@@ -16,7 +16,7 @@ func TestResourceName(t *testing.T) {
 
 func TestTypes(t *testing.T) {
 	assert := assert.New(t)
-	assert.Implements((*handlers.ResourceHandler)(nil), &LxcHandler{})
+	// assert.Implements((*handlers.InstanceResourceHandler)(nil), &LxcHandler{})
```
unakatsuo: Why is this commented?

handlers/vm/null/null_test.go
```diff
@@ -14,6 +14,6 @@ func TestResourceName(t *testing.T) {
 
 func TestTypes(t *testing.T) {
 	assert := assert.New(t)
-	assert.Implements((*handlers.ResourceHandler)(nil), &NullHandler{})
+	// assert.Implements((*handlers.InstanceResourceHandler)(nil), &NullHandler{})
```
unakatsuo: Why is this commented?

handlers/vm/qemu/qemu_test.go
```diff
@@ -16,7 +16,7 @@ func TestResourceName(t *testing.T) {
 
 func TestTypes(t *testing.T) {
 	assert := assert.New(t)
-	assert.Implements((*handlers.ResourceHandler)(nil), &QemuHandler{})
+	// assert.Implements((*handlers.InstanceResourceHandler)(nil), &QemuHandler{})
```
unakatsuo: Why is this commented?

scheduler/mesos_test.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+var (
+	mesosOffer *mesos.Offer
+)
+
+func init() {
+	mesosOffer = &mesos.Offer{
+		Id: &mesos.OfferID{
+			Value:            sP("d39c0128-4822-49a0-9fab-640fba518d53-O590"),
```
unakatsuo: ``golang/protobuf`` provides conversion utilities.

```
proto.String("xxxxxx")
```

https://github.com/golang/protobuf/blob/master/proto/lib.go

scheduler/schedule.go
```diff
@@ -0,0 +1,93 @@
...
+func RegisterInstanceScheduleHandler(name string, i handlers.InstanceScheduleHandler) error {
+	if _, exists := instanceSchedulerHandlers[name]; exists {
+		return fmt.Errorf("Duplicated name for instance schedule handler: %s", name)
+	}
+	instanceSchedulerHandlers[name] = i
+	return nil
+}
+
+type Schedule struct {
+	*sync.Mutex
```
unakatsuo: No need to expose Lock()/Unlock() methods. And also ``sync.RWMutex`` is better .

https://golang.org/pkg/sync/#RWMutex



```
m sync.RWMutex
```

scheduler/schedule.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// TODO write definition of openvdc offer (interface | struct)
+func (s *Schedule) StoreOffer(offer *model.VDCOffer) {
+	schedule.Lock()
+	defer schedule.Unlock()
+	schedule.storedOffers[offer.SlaveID] = offer
+}
+
+func (s *Schedule) Assign(inst *model.Instance) error {
```
unakatsuo: Missing to take mutex lock.

unakatsuo (changes requested):

handlers/vm/lxc/lxc_test.go
itouri: A little time ago,  InstanceResourceHandler was registered by each Handler. But now InstanceResourceHandler is not used by any handlers, because each Scheduler registering instanceSchedulerHandler to global variable of scheduler/schedule.go instead of InstanceResourceHandler. So each Hander need not implement InstanceResourceHandler. I deleted this comment.

handlers/vm/null/null_test.go
itouri: Same as above.

handlers/vm/qemu/qemu_test.go
itouri: Same as above.

scheduler/mesos_test.go
itouri: Modified to using proto library for convert to pointer.

scheduler/schedule.go
itouri: Modified mutex from sync.Mutex to sync.RWMutex and turn to private variable.

scheduler/schedule.go
itouri: Added RLock/RUnlock next to schedule.storedOffers's "for range" function.

PR #212, opened 2017-12-04
Terraform provider

Metallion:
Until now I had never thought about merging the terraform provider branch because I figured it wasn't finished. It doesn't implement all of OpenVDC's features yet.

Now I figured it's already being used in another project and it's working for what it's being used in there. Therefore it's better to get it merged after all. That way we can do proper branched development on the terraform provider too.


Metallion:
I see the unit tests for the terraform provider are currently broken. I'll re-open this PR when i get around to fixing them.

PR #213, opened 2017-12-04
Force destroy

b0r6:
```./openvdc destroy --force i-000000001```

This skips the state validation when trying to terminate the instance, making it possible to terminate them even though they might be stuck as RUNNING, for example.

### CODE REVIEW

**Review by Metallion — CHANGES_REQUESTED** [review]
Still needs test code. If it's possible to do just unit tests, I think that should be enough for a feature like this. If it's not possible, then maybe a small acceptance test.

Btw, would it also be possible to force other states? For example set an instance to RUNNING when it's stuck in STARTED because of a zookeeper connection error but the container/vm did come up correctly? (Doesn't necessarily need to be added to this PR)

**Comment by b0r6** [comment]
@Metallion 

> Still needs test code. If it's possible to do just unit tests, I think that should be enough for a feature like this. If it's not possible, then maybe a small acceptance test.
> 
> Btw, would it also be possible to force other states? For example set an instance to RUNNING when it's stuck in STARTED because of a zookeeper connection error but the container/vm did come up correctly? (Doesn't necessarily need to be added to this PR)

Alright, I'll add tests.

Yeah that wouldn't be hard to do I guess. Do you mean something like ```./openvdc forcestate RUNNING i-000000001```

**Comment by Metallion** [comment]
Something like that, yes. Though I think it would be more intuitive to have the instance id before the state. The openvdc wait command does that too.

```
openvdc forcestate i-000000001 RUNNING
```

**Review by Metallion — APPROVED** [review]

**Comment by Metallion** [comment]
Looks good to me but if possible, i'd still like to get somebody else to review the code.

File: cmd/openvdc/cmd/destroy.go [hunk 1]
```diff
@@ -10,8 +10,12 @@ import (
 	"google.golang.org/grpc"
 )
 
+func init() {
+	destroyCmd.Flags().Bool("force", false, "Force destroy instance, ignoring states")
+}
+
 var destroyCmd = &cobra.Command{
-	Use:   "destroy [Instance ID]",
+	Use:   "destroy [Instance ID] [options] [--] [commands]",
```
unakatsuo: Does ``openvdc destroy`` accept extra command args?

ci/citest/acceptance-test/tests/force_destroy_test.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+func TestForceDestroy(t *testing.T) {
+        stdout, _ := RunCmdAndReportFail(t, "openvdc", "run", "centos/7/lxc", `{"interfaces":[{"type":"veth"}], "node_groups":["linuxbr"]}`)
+        instance_id := strings.TrimSpace(stdout.String())
+
+        WaitInstance(t, 5*time.Minute, instance_id, "RUNNING", []string{"QUEUED", "STARTING"})
+
+	//Simulate stuck state
+	RunCmdAndReportFail(t, "openvdc", "force-state", instance_id, "starting")
```
unakatsuo: Missing gofmt.

unakatsuo (changes requested):

unakatsuo (approved):

PR #214, opened 2017-12-08
Feature force state

b0r6:
```./openvdc force-state i-0000000001 running```

Allowing states to be forcefully set via the client might prove useful during development and debugging.

### CODE REVIEW

**Comment by Metallion** [comment]
Looks good to me. Gonna try it for myself a bit first though. :) Also one minor thing. I find `forcestate` as one word a little hard to read. Think we could get a hyphen in there? `force-state`

```
./openvdc force-state i-0000000001 running
```

Metallion (approved):

PR #215, opened 2017-12-14
Feature esxi vcenter

toros11:
This PR adds following features to the esxi driver

- Ability to use vCenter for instance creating
- Set the datastore location/base image name in json template (vmkfs cloning)
- Use ovf templates as the base for which to create new instances (vCenter)
- Use inbuilt clone feature (vCenter)
- Add nic/set network id
- Add floppy device used for metadata image if not present


pkg/conf/executor.toml
```diff
@@ -13,8 +13,11 @@
 # esxi-datacenter = "ha-datacenter"
 # esxi-insecure = true
 # esxi-host-sshkey = ""
+# esxi-easy-clone = false
```
b0r6: It seems like the options ```'esxi-easy-clone'``` and ```'esxi-basevm-datastore'``` isn't needed anymore.

hypervisor/esxi/esxi.go
```diff
@@ -97,6 +86,7 @@ func (p *EsxiHypervisorProvider) Name() string {
 func init() {
 	hypervisor.RegisterProvider("esxi", &EsxiHypervisorProvider{})
 	viper.SetDefault("hypervisor.esxi-insecure", true)
+	viper.SetDefault("hypervisor.vCenter", false)
```
unakatsuo: This parameter breaks the existing naming convention. It would be better to start with ``esxi-`` prefix.

hypervisor/esxi/esxi.go
```diff
@@ -216,6 +247,7 @@ func runCmd(cmd string, args []string) error {
 	return nil
 }
 
+var ErrApiRequest = errors.New("Failed api requiest")
```
unakatsuo: typo...

hypervisor/esxi/esxi.go
```diff
@@ -208,6 +204,41 @@ func join(separator byte, args ...string) string {
 	return buf.String()
 }
 
+func captureStdout(fn func() error) ([]byte, error) {
+	r, w, err := os.Pipe()
```
unakatsuo, line 208: ``r`` and ``w`` should be closed properly until it returns from this function.

unakatsuo (changes requested):

hypervisor/esxi/esxi.go
toros11: vCenter doesn't have anything to do with esxi though? It's just the endpoint.

Maybe it makes more sense to make a separate optional config field in the .toml file?

```
[vcenter]
enabled = true/false
```

b0r6 (approved):

Metallion:
Merging since @b0r6 has approved, @unakatsuo's not available for review and there is a commit addressing his comment.

PR #218, opened 2017-12-18
Terraform provider

akirapop:
This seems to be in working order. It would be nice if terraform would allow an information message ("instance already terminated. ignoring.") but it's not clear how to do that.


Metallion:
Can't be merged yet since it doesn't pass the CI but I'll have a look at the code.

terraform/openvdc/provider.go
```diff
@@ -0,0 +1,71 @@
...
+
+func CheckInstanceTerminated(cmdStdOut *bytes.Buffer) (bool, error) {
+	status, err := ParseJson(cmdStdOut)
+
+	if err != nil {
+		return false, err
+	}
+
+	if strings.Compare(status, "FAILED") == 0 {
+		return false, fmt.Errorf("Instance is in FAILED state!")
```
Metallion: Unless I'm missing something this in combination with [line 100 of resouce_openvdc_instance.go](https://github.com/axsh/openvdc/pull/218/files#diff-654da0062ce0dead0ec37848cdcc86beR100) will still cause `terraform destroy` to fail if the instance is already in FAILED state.

Metallion (changes requested):
The issue currently on the CI was fixed locally by running `go get <github url>/terraform`, right?

If that's the case, how about we just add that line to build.go? We already have [similar commands](https://github.com/axsh/openvdc/blob/master/build.go#L155-L160) in there for go-binddata.

terraform/openvdc/provider.go
akirapop: Right. I had thought that this was wanted. (If not, there is no point in checking for FAILED, RUNNING, etc.)

terraform/openvdc/provider.go
```diff
@@ -0,0 +1,71 @@
...
+
+	return &stdout, &stderr, err
+}
+
+func ParseJson(cmdStdOut *bytes.Buffer) (string, error) {
+	var parsed map[string]interface{}
+
+	err := json.Unmarshal(cmdStdOut.Bytes(), &parsed)
+	if err != nil {
+		fmt.Println("Uh-oh! You had an error!")
```
unakatsuo: It's obvious just to return error object. No need for message to stdout.

b0r6:
@Metallion Wouldn't the ci error get fixed if the terraform packages were added to vendor.json?

terraform/openvdc/provider.go
Metallion: The thing is that if an instance failed to come up and thus is in FAILED state, terraform should be able to retry creating it. Running `terraform destroy` to clean up faulty resources and then running `terraform apply` again is the most intuitive way to do this.

That's why the destroy function in the provider should return nil when an instance is in FAILED state.  :) Returning nil tells terraform that the resource has successfully been destroyed and thus it's removed from the .tfstate file.

If trying to destroy a FAILED instance returns an error, the user will have to force remove it from terraform's .tfstate file in order to try again.

unakatsuo (changes requested):

akirapop:
@b0r6 There are versioning issues going on, IIRC.  (The terraform builder requires go v. 1.9-ish. The latest govendor seems to fetch 1.8x.)

Metallion:
@b0r6 We tried that. The problem is that it also pulls in a huge number of dependencies. Basically it pulls in libraries for all the cloud platform that terraform has built-in providers for. We don't need all that just to compile our own provider.

ci/citest/unit-tests/unit-tests.sh
```diff
@@ -46,5 +46,6 @@ docker cp . "${CID}:/var/tmp/go/src/github.com/axsh/openvdc"
 
 ## Run unit tests
 docker exec $CID /bin/bash -c "/usr/bin/env"
-docker exec $CID /bin/bash -c "cd /var/tmp/go/src/github.com/axsh/openvdc;  govendor sync"
+#docker exec $CID /bin/bash -c "cd /var/tmp/go/src/github.com/axsh/openvdc;  govendor sync"
```
Metallion: I assume this was accidentally committed? :)

akirapop: ? I'm confused. That's not what I've got (and it's not what's just run - and failed - on the ci).
Pushing soon.

Metallion:
I know I was the one who suggested adding `go get terraform` to the build script but I didn't realize this needed to go in the unit test script for the CI to pass. Since dependencies should be managed in one place, I felt like it shouldn't go there.

Then I had a look at the actual tests being done for the terraform provider. Currently it's just an `InternalValidate()` and I figured it is ok to skip this on the CI for now. Therefore I have added a build constraint to all terraform files, excluding them from the standard build process.

Metallion (approved):

terraform/openvdc/provider.go
toros11: Additionally this function name is a bit misleading, it only returns a hardcoded parameter.

What about renaming it `ParseState()` or something similar?

terraform/openvdc/provider.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func CheckInstanceTerminated(cmdStdOut *bytes.Buffer) (bool, error) {
+	status, err := ParseJson(cmdStdOut)
+
+	if err != nil {
+		return false, err
+	}
+
+	if strings.Compare(status, "TERMINATED") == 0 {
```
unakatsuo:
```
return (status == "TERMINATED"), nil
```
is enough.

unakatsuo (changes requested):

unakatsuo (approved):

PR #219, opened 2017-12-19
Add crashed node immediately

Metallion:
It takes some time mesos to register a crash. Restarting mesos-slave before first waiting until a crash was registered causes the crash recovery feature to not kick in at all.

However, when mesos-slave is stopped, it does send `TASK_LOST` events with `SLAVE_DISCONNECTED` as the reason to openvdc-scheduler. I hooked into those events to add the crashed node immediately when mesos-slave goes down.

Of course it is still possible for mesos-slave to go down without sending the event if for example power is suddenly lost to the host. In that case the old behaviour of adding the crashed node after some time has passed still works.

### DIFF
--- a/scheduler/mesos.go
+++ b/scheduler/mesos.go
@@ -503,6 +503,11 @@ func (sched *VDCScheduler) StatusUpdate(driver sched.SchedulerDriver, status *me
 		driver.ReviveOffers()
 	}
 
+	if status.GetState() == mesos.TaskState_TASK_LOST &&
+		status.GetReason() == mesos.TaskStatus_REASON_SLAVE_DISCONNECTED {
+		sched.SlaveLost(nil, status.GetSlaveId())
+	}
+
 	if status.GetState() == mesos.TaskState_TASK_LOST ||
 		status.GetState() == mesos.TaskState_TASK_ERROR ||
 		status.GetState() == mesos.TaskState_TASK_FAILED ||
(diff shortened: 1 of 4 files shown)


scheduler/mesos.go
```diff
@@ -523,13 +528,22 @@ func (sched *VDCScheduler) FrameworkMessage(_ sched.SchedulerDriver, eid *mesos.
...
-	agentMesosID := *sid.Value
-
 	ctx, err := model.Connect(context.Background(), sched.zkAddr)
 	if err != nil {
 		log.WithError(err).Error("Failed model.Connect")
 	}
 
+	agentMesosID := *sid.Value
+
+	crashed_node, err := model.CrashedNodes(ctx).FindByAgentMesosIDNotReconnected(agentMesosID)
```
b0r6, line 538: Couldn't the same result already be achieved by using the function ```FindByAgentMesosID``` and then just checking GetReconnected?
```
if crashed_node != nil  && crashed_node.GetReconnected() == false {
		// Node already added to crashed nodes. Nothing to do.
    return
}
```

Metallion: If a node crashes and reconnects multiple times, it will get a new entry in zookeeper for every crash. `FindByAgentMesosID` only returns the first entry found so that one might have Reconnected set to true while there's still another one with Reconnected set to false.

Thus the above example would cause a duplicate to get added to crashes nodes.

b0r6: Ah yeah, never mind then, I had forgotten that it only returns the first value ^^

Metallion: So, is it ok for merge then? :)

b0r6: Looks ok to me!

b0r6 (approved):

PR #223, opened 2018-10-05
Terraform provider update

toros11:
* Allow to set gateway from terraform resource
* Add provider config to set the api endpoint for openvdc-scheuler


Metallion (approved):
:+1:

